PR #154, opened 2016-02-15
Removing Scenario::getEndTimeNode() and removing TimeProcess getStartState() and getEndState() accessors

theod:
as proposed in https://github.com/OSSIA/API/issues/18 I have removed Scenario::getEndTimeNode.

but in the process I've noticed there were also useless accessors for TimeProcess (getStartState() and getEndState()). Those states were added behind the scene to start and end time events when a TimeProcess was added to a TimeCosntraint. but it seems those states was never used in i-score and I'm pretty sure they are useless as it is to the API user to decide what to put into the start and end events.

what do you think @jcelerier and @nvuaille ?
let me know if it is ok to merge feature/api_clean ?

### DIFF
--- a/base/plugins/iscore-plugin-js/JS/JSProcess.cpp
+++ b/base/plugins/iscore-plugin-js/JS/JSProcess.cpp
@@ -20,9 +20,7 @@ namespace Executor
 {
 ProcessExecutor::ProcessExecutor(
         const Explorer::DeviceDocumentPlugin& devices):
-    m_devices{devices.list()},
-    m_start{OSSIA::State::create()},
-    m_end{OSSIA::State::create()}
+    m_devices{devices.list()}
 {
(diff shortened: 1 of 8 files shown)


jcelerier:
Yes, they were just here because they had to be implemented but I don't remember using them at all. Thanks !

jcelerier:
However I am sorry but changing this will certainly break the save format (since an "end" event will not be expected anymore). I will try to make it so that a "loaded" scenario will just have its end event be a "normal" event at the end.

theod:
ok fine !
do I need to wait before the merge or is it possible to merge and take care of this later ?

jcelerier:
I am currently doing the merge but there are a bunch of stuff tochange in i-score.

theod:
thanks !

PR #832, opened 2018-09-14
Wiimote Integration

aliefhooghe:
Wiimote Integration in score, using the Libossia Wiimote protocol. One device only can be instantiate at a time. A device and can handle 4 wiimote (One wiimote per branch). Therefore settings are empty and thus are easily serialized. A ProgressDialog is used in order to make the user wait while the connection. It take a few second. 

I still have the deadlock Problem when the device is removed after having shown the parameter tree.  I will open an issue for more detailed information.


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "API"]
 	path = API
-	url = https://github.com/OSSIA/API
+	url = https://github.com/aliefhooghe/libossia.git
```
jcelerier: please revert this

aliefhooghe: It is done

jcelerier:
thanks !

PR #837, opened 2018-09-24
Artnet DMX integration

aliefhooghe:
The DMX implementation being basic, the only setting is the device name and there is no serialization.
It is a base work for future enhancement.


jcelerier:
@aliefhooghe do you think that you could do the following improvements : 

- allow to choose the universe in the artnet settings widget
- add an IP chooser (maybe with the default being the broadcast address?)
- in the device, just name the channels 1, 2, 3, ... instead of Channel-1, Channl-2, ... more characters means more potential errors

thanks !

aliefhooghe:
I will look that !
- IP I can easily add a 'preferred' IP that will be chosen among  
available Artnet ip (this is actualy what artnet do). I don't know if  
it will be easy to detect failure (ip not found)
- I will check for universe and rename channels !

Quoting Jean-Michaël Celerier <notifications@github.com>:

PR #848, opened 2018-10-07
Update .gitmodules

Dhanu-29:
Line 3 was changed


jcelerier:
hmm sorry but why would we change where the submodule points to ?

PR #856, opened 2018-10-11
necessary deletion

thibaudk:
I was had a little error in the last build

```
score/base/lib/core/document/Document.cpp:29:58: error: no ‘score::DocumentContext score::DocumentContext::fromDocument(score::Document&)’ member function declared in class ‘score::DocumentContext’
 DocumentContext DocumentContext::fromDocument(Document& d)
``` 

And it seemed that these couple lines where doubling up with the following few.
builds fine on Ubuntu 18.04 with this tiny deletion.


jcelerier:
thanks !

PR #860, opened 2018-10-25
MIDI Pitch in Midi categorie

thibaudk:
Tiniest clean-up.
Midi Pitch had it's own metadata category.
Here it is in the same category as all other midi processes.


jcelerier:
thanks !

PR #863, opened 2018-11-08
Joystick & wiimote support as default

thibaudk:
Since Joystick and wiimote support where actually more tested than atnet (witch already builds by default), maybe it would be good to add them to the default config.

If it is only to make sure that it continues to build with future commits.

This is just a suggestion.


thibaudk:
This pr requires the addition of SDL2 and libbluetooth as dependencies.

thibaudk:
After testing, a problem appeared with wiiuse shared library.
For some reason, score can't find it and won't lauch without.

Score can be built with wiimote support & without wiiuse shared library.
It lauches, the wiimote Device can be created, the tree looks correct but no values are updated

thibaudk:
close for now as both are disabled

PR #993, opened 2020-01-17
Stable

thibaudk:
another attempt


thibaudk:
fixed through 37b4401

PR #1257, opened 2021-07-22
add of compute shader buffers for particles

Afourcadet:


jcelerier:
I think that this should go in the particle addon's node class actually, there's no particular reason for this to be in score

PR #1258, opened 2021-07-22
Obey [style:knob] UI syntax for faust

thibaudk:
Very basic.
Further improvement would involve trimming down repetition in the code,
as well as Obeying as much of Faust UI syntax as possible.

Thanks a million for the assist @jcelerier


jcelerier (approved):
thanks !

PR #1269, opened 2021-08-02
added XYZSlider widget

thibaudk:
This is basically a mix between the HSVSlider and the XYSlider.
The "final" piece I think would be for the ```bindVec2Domain``` to extend to vec3f.
https://github.com/ossia/score/blob/81d54f78790ce4805b8d074b40fe8a06cf986171/src/plugins/score-lib-process/Process/Dataflow/ControlWidgetDomains.hpp#L197
here, vec2f are hard coded, but maybe they could be replaced by another template argument ?
Or maybe, the type could be deduced from one of the other template argument ?

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -259,6 +259,7 @@ set(HEADERS
     "${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/widgets/QGraphicsSliderBase.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/widgets/QGraphicsToggle.hpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/widgets/QGraphicsXYChooser.hpp"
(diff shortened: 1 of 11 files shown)


thibaudk:
I am thinking as well, maybe I should have started by inheriting from the XYSlider ? 
Would it be better ?

jcelerier:
nope definitely no inheritance, those are 100% unrelated

thibaudk:
I was thinking about it because it would solve the issue of ```bindVec2Domain```
but I guess a lot of things would have to be overridden

jcelerier:
just make a bindVec3domain like there's a bindIntDomain, bindFloatDomain, etc... the ideal solution would be instead to have a template that handles all this but it's a tricky one to write to handle all cases correctly, and would ultimately make code harder to read / debug

thibaudk:
Thanks for the tip ! It now all seem to work as it should after a few tests.

jcelerier:
pretty sweet, thanks !!

PR #1283, opened 2021-08-14
add QCodeEditor

thibaudk:
as mentioned in 
https://github.com/ossia/sdk/issues/1

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -20,6 +20,7 @@ file(GENERATE
 
 if(SCORE_INSTALL_HEADERS)
   install(FILES
+        ${3RDPARTY_FOLDER}/QCodeEditor
         ${CMAKE_CURRENT_BINARY_DIR}/zipdownloader_export.h
         ${CMAKE_CURRENT_BINARY_DIR}/rnd_export.h
         DESTINATION include/score


jcelerier:
sorry, build was broken, can you rebase on top of master to check the current status? 

```
git checkout feature/install_headers
git fetch
git rebase <remote name of the score repo>/master
git push --force
```

If you don't have the score repo as remote (which can be checked with `git remote -v`): run `git remote add ossia git@github.com:ossia/score` and then use `git rebase ossia/master`

thibaudk:
done ! thanks
let's see how it goes

I always have ossia as a remote by the way.
I like to have it for fetching and have my branch for pushing, I find it easier
```
git remote -v
origin	git@github.com:ossia/score.git (fetch)
origin	git@github.com:thibaudk/score.git (push)
```

thibaudk:
mmmhh,

i still get 

https://dev.azure.com/ossia/libossia/_build/results?buildId=1844&view=logs&j=9eac21c1-0c13-5808-5e12-e11dc93d3cd2&t=dd780cb1-b74f-537b-c827-058d2f62a320&l=7661

src/lib/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ file(GENERATE
 
 if(SCORE_INSTALL_HEADERS)
   install(FILES
+        ${3RDPARTY_FOLDER}/QCodeEditor
```
jcelerier, line 23: I don't think that this will work, did you check what gets copied if you make a build with 

     cmake <score folder> -GNinja -DSCORE_INSTALL_HEADERS=1 -DCMAKE_INSTALL_PREFIX=/tmp/installed -DDEPLOYMENT_BUILD=1 
     cmake --build .
     cmake --build . --target install
     
?

thibaudk:
I didn't, didn't know that option.
let me see

src/lib/CMakeLists.txt
thibaudk: ok so this works with the include ```${3RDPARTY_FOLDER}/QCodeEditor/include/internal/QCodeEditor.hpp```
I wonder if I need to include all of QCodeEditor files tho ?

jcelerier:
good question...
I wonder if instead of doing this, QCodeEditor shouldn't be hidden as an implementation detail in the score library.
What functions of it do you need ? Do you just call "new QCodeEditor" ?

thibaudk:
actually i call ```Process::createScriptWidget("JS");```
that returns a new QCodeEditor

jcelerier:
okay, can you try with the latest commit https://github.com/ossia/score/commit/cdc614e2bc80e71d0b35be4bb3467197d5142de3 and with replacing all QCodeEditor by QTextEdit (its base class) in your code ?

thibaudk:
Thanks @jcelerier after chaging QCodeEditor for QTextEdit in the addon, I still get another error related to missing includes
https://github.com/thibaudk/score-addon-stdIO/runs/3334225585?check_suite_focus=true#step:6:329

jcelerier:
you should not need to include  score/src/plugins/score-plugin-protocols/Protocols/Serial/SerialSpecificSettings.hpp (or anything from <Protocols/Serial/...>) in your own code

thibaudk:
my bad !
i inclued ```Protocols/ProtocolLibrary.hpp``` but I only needed ```ossia-qt/js_utilities.hpp```
from it for the ```  qmlRegisterType<stdIO::StdIO>("Ossia", 1, 0, "StdIO");```
Thanks for the heads up

thibaudk:
after fixing this last error, the jit build still gets cancelled,complainig that the addon doesn't have anny cpp files
https://github.com/thibaudk/score-addon-stdIO/runs/3351665033?check_suite_focus=true#step:7:15

thibaudk:
ok it is probalby because i didn't create a release

jcelerier:
no don't worry about that, i have to fix it
if your plug-in works we're good :)

PR #1287, opened 2021-08-18
Feature/wasm remote

Toine-ddt:


Toine-ddt:
@jcelerier

.gitmodules
```diff
@@ -44,3 +44,6 @@
 [submodule "3rdparty/hap"]
 	path = 3rdparty/hap
 	url = https://github.com/Vidvox/hap
+[submodule "3rdparty/qml-remote"]
```
jcelerier, line 47: No I don't think we want the qml remote repository like that

jcelerier:
@thibaudk for this it should be a package shipped in ossia/score-packages so that people who don't use it don't pay the cost. Then, we should have a simple boost.beast HTTP server to serve it, for instance in /remote.

jcelerier:
@thibaudk also I merged the release and the support for passing the IP over an environment variable that @Toine-ddt had done : https://github.com/ossia/qml-remote  ; look how to read it here if it's not already done

thibaudk:
> @thibaudk also I merged the release and the support for passing the IP over an environment variable that @Toine-ddt had done : https://github.com/ossia/qml-remote ; look how to read it here if it's not already done

Ok perfect. That would be very useful.
I'll start turning this into its own plugin / package.
Exciting stuff !

jcelerier:
correct pr is #1288 closing this one

PR #1296, opened 2021-09-10
fix #1295

### LINKED ISSUE #1295: Units are not saved in the adresse of triggers
when setting up a trigger to react to particular values from a parameter, say the Y value of the of a cart2D, the access needs to be set in inspector as ```Device:/adress/of/the/parameter@[position.cart2D.y]```.
Yet, lately, the unit and extended types are not reloaded in the scenario.
The address of the parameter either appears on it's own, without the unit, or the unit is replaces by an integer and lacks the ```@``` character ```Device:/adress/of/the/parameter[1]```.
Furthermore, after entering a proposer unit, any other changes to the model will temper withe the unit in the same way. It is directly visible in the score and in the .scorejson file 

The ```@[N]``` syntax, enabling the access of a specific value inside an array is not valid anymore either.

thibaudk:
this seem to do the trick, 
it was just a forgotten '@' character

### DIFF
--- a/src/plugins/score-lib-state/State/Relation.cpp
+++ b/src/plugins/score-lib-state/State/Relation.cpp
@@ -38,7 +38,7 @@ QString State::toString(const State::RelationMember& m)
       auto addr = acc.address.toString();
       for (auto val : acc.qualifiers.get().accessors)
       {
-        addr += QString("[%1]").arg(val);
+        addr += QString("@[%1]").arg(val);
       }
       return "%" + addr + "%";
     }


jcelerier (commented):
arf I'm sorry this fix is incorrect :( if there are multiple accessors it will be

   foo:/bar@[1]@[2]

while it should be 

   foo:/bar@[1][2]

so there need to be an `if (!accessors.empty()) addr += '@';` instead before

thibaudk:
In 4 years using score, multiple accessors still never occurred to me.
you can find the changes in #1299
Thanks !

jcelerier:
thanks for the fix :)

PR #1308, opened 2021-10-02
Wip noise filter

thibaudk:
please refer to https://github.com/ossia/libossia/pull/746

### DIFF
--- a/src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
+++ b/src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
@@ -65,8 +65,7 @@ struct Node
 
         vis(NoiseFilter* parent)
             : nFilt{parent}
-        {
-        }
+        {}
       };
 
       try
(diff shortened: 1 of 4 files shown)


src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
```diff
@@ -0,0 +1,340 @@
...
+
+        vis(NoiseFilter* parent)
+          : nFilt{parent}
+        {}
+      };
+
+      try {
+        return ossia::apply(vis{this}, val.v);
+      } catch(std::exception& e) {
+        std::cerr << e.what() << std::endl;
```
jcelerier: if you can, use `ossia::logger().error("{}", e.what());` instead so that it shows in the message log

src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
```diff
@@ -0,0 +1,340 @@
...
+      } catch(...) {
+        std::cerr << "error" << std::endl;
+      }
+
+      return val;
+    }
+
+    void setAmount(const float& amt)
```
jcelerier: when passing arguments of int, float, double type always use the value directly instead of a const reference, it's faster

src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
```diff
@@ -0,0 +1,340 @@
...
+        return value;
+    }
+
+  private :
+    std::string prevType{"OneEuro"};
+    float prevAmount{0.1}, prevFreq{120.}, prevCutoff{1.}, prevBeta{1.};
+
+    NoiseFilter nf{};
+
+    type GetFilterType(std::string_view str) noexcept
```
jcelerier, line 228: can be marked static

jcelerier (commented):
looks pretty good, can't wait to try it ^^

src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
thibaudk: done with 0931d37

src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
thibaudk: done with 7d7922bc357fb7b71b376985d713529291013733

src/plugins/score-plugin-fx/Fx/ValueFilter.hpp
thibaudk: done with 7d7922bc357fb7b71b376985d713529291013733

thibaudk:
Thanks a lot for the review !
Unfortunately, 2 major issue remain:
The first and most important one being that it does not filter. For some reason, the input value is now always returned with no apparent modification.
The second is that changing type seem to change the index of the variant, but not actually change what filter type is being used.

Will keep investigating fuhrer, but something here is a bit weird.

jcelerier:
how's that going ? do you want me to look into it ? I think it would be good to have it in v3

thibaudk:
Mostly there, sorry, had to leave it for a bit.
Still remaining is the controls to the low pass filter as well as the rate and quantization.

thibaudk:
done with bcef37db9a266400e5385c958008eb2ec301f493

PR #1340, opened 2021-12-15
Added Midi Note deletion using the Delete Key for Piano Roll plugin

RobinMontferme:
Fixed Midi note removal through Delete Key


jcelerier (approved):
looks great, thanks !

PR #1341, opened 2021-12-15
Attempt at improving UX with the package manager

thibaudk:
* show remote packages size
* reshape packages Tableview to display version numbers and leave more room for descriptions
* "open Link" button
* separate table and buttons, as the "open Link" works for both tabs. The progress was also moved to the bottom of the view to minimize resizing.   
* PackagesModel class to avoid redundancy
* The signal sent by https://github.com/ossia/score/blob/master/src/plugins/score-plugin-packagemanager/PackageManager/PluginItemModel.cpp#L51 never actually ended up registering the add-on dynamically. I suspect the reason to be that the signal triggers when the new package folder is created, but the addon.json file might not yet be downloaded. Instead, I propose for registerAddon to be called after the package was successfully installed https://github.com/thibaudk/score/blob/wip/pkgMng/src/plugins/score-plugin-packagemanager/PackageManager/View.cpp#L444

### DIFF
--- a/src/plugins/score-plugin-library/Library/LibrarySettings.cpp
+++ b/src/plugins/score-plugin-library/Library/LibrarySettings.cpp
@@ -113,6 +113,9 @@ void Model::firstTimeLibraryDownload()
 
             rescanLibrary();
           },
+      [] (qint64 bytesRecieved, qint64 bytesTotal) {
+        qDebug() << (((bytesRecieved / 1024.) / (bytesTotal / 1024.)) * 100)
+                 << "% downloaded"; },
           [] {});
     }
   }
--- a/src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
+++ b/src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
@@ -4,11 +4,6 @@
 
 #include "Presenter.hpp"
 
-#include <Library/LibrarySettings.hpp>
-
-#include <score/application/ApplicationContext.hpp>
-#include <score/plugins/settingsdelegate/SettingsDelegateView.hpp>
-
 #include <QBuffer>
 #include <QDir>
 #include <QFile>
@@ -22,6 +17,7 @@
 #include <QNetworkRequest>
 #include <QStandardPaths>
 #include <QTemporaryFile>
+#include <QDesktopServices>
 
 #include <PackageManager/FileDownloader.hpp>
 #include <score_git_info.hpp>
@@ -31,7 +27,6 @@
 W_OBJECT_IMPL(PM::PluginSettingsView)
 namespace PM
 {
-
 namespace zip_helper
 {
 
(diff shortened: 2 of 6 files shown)


jcelerier:
~why has LocalPackageModel been renamed into packageModel ?~ ah I understand, cool !

can you just rename it with the proper capitalization ? (PackagesModel)

thibaudk:
sure ! here it is
Still a good few things to improve

thibaudk:
Still remaining :
- [X] Show space left on disc
- [x] Uninstall option
- [x] Make progress bar work
- [ ] Update option
- [ ] Only show packages that have not yet been installed when Browsing
- [ ] multiple selection (optional)

jcelerier:
checking it tomorrow :)

thibaudk:
No rush, still a good few things to do.

jcelerier:
ok ping me when you think it's good :)

thibaudk:
This last commit requires the changes proposed in another pr for the zipdownloader third party
https://github.com/jcelerier/zipdownloader/pull/1

The additional tasks for this Pr now seem quite optional.
They could be converted to issues about further improvements.

The only thing left I would be worried about here are some inconsistencies with the progress bar.
After first getting them, I then was not able to reproduce them.
Feedback would be greatly appreciated !

thibaudk:
> This last commit requires the changes proposed in another pr for the zipdownloader third party [jcelerier/zipdownloader#1](https://github.com/jcelerier/zipdownloader/pull/1)

This is now done and rebased.

src/plugins/score-plugin-library/Library/LibrarySettings.cpp
```diff
@@ -113,6 +113,9 @@ void Model::firstTimeLibraryDownload()
 
             rescanLibrary();
           },
+      [] (qint64 bytesRecieved, qint64 bytesTotal) {
```
jcelerier, line 116: what remains is to put that in a label right ?

thibaudk: If you wish to show the progress of the library download. Although I thought that just letting the download happen in the background was less distracting. Where would that label go ? In the popup ?

jcelerier: hmmm maybe we could put it at the bottom of the "info" panel at the bottom right

jcelerier: (not even showing the value, just a progress bar for each download)

thibaudk: Sure !
Should this be for another PR or this one ?

jcelerier: yes definitely, let's keep this one as is

src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
```diff
@@ -213,155 +239,224 @@ void PluginSettingsView::on_message(QNetworkReply* rep)
...
-#else
+    #else
       "linux"
-#endif
+    #endif
       ;
 
   const QString sdk_path{
-      score::AppContext().settings<Library::Settings::Model>().getSDKPath() + "/" + version};
+    score::AppContext().settings<Library::Settings::Model>().getSDKPath() + '/' + SCORE_TAG_NO_V};
```
jcelerier, line 406: better to keep using version here, it saves a string -> qstring conversion

src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
```diff
@@ -213,155 +239,224 @@ void PluginSettingsView::on_message(QNetworkReply* rep)
...
+  if (m_install->isVisible())
+    return static_cast<PackagesModel*>(m_remoteAddons->model());
+  else
+    return static_cast<PackagesModel*>(m_addonsOnSystem->model());
+}
 
-  auto rows = m_remoteAddons->selectionModel()->selectedRows(0);
-  if (rows.empty())
-    return;
+int PluginSettingsView::get_current_row(const QTableView* t = nullptr)
```
jcelerier: please use camelCase for methods in score if it's mostly Qt-ish things !

src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
```diff
@@ -213,155 +239,224 @@ void PluginSettingsView::on_message(QNetworkReply* rep)
...
+{
+  QModelIndexList rows{};
+
+  if (t)
```
jcelerier, line 263: { 
}

src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
```diff
@@ -213,155 +239,224 @@ void PluginSettingsView::on_message(QNetworkReply* rep)
...
+  if (rows.isEmpty())
+    return -1;
 
-  auto num = rows.first().row();
-  SCORE_ASSERT(int(remotePlugins.addons().size()) > num);
-  auto& addon = remotePlugins.addons().at(num);
+  return rows.first().row();
+}
+
+Package PluginSettingsView::selected_Package(const PackagesModel* model, int row)
```
jcelerier: selectedPackage

src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
```diff
@@ -213,155 +239,224 @@ void PluginSettingsView::on_message(QNetworkReply* rep)
...
+{
+  if (row == -1)
+    return {};
+
+  SCORE_ASSERT(int(model->addons().size()) > row);
+
+  return model->addons().at(row);
+}
+
+void PluginSettingsView::open_link()
```
jcelerier: openLink

src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
```diff
@@ -213,155 +239,224 @@ void PluginSettingsView::on_message(QNetworkReply* rep)
...
+{
+  auto addon = selected_Package(get_current_model(), get_current_row());
```
jcelerier: etc etc

jcelerier: (also const auto& here would be nicer)

src/plugins/score-plugin-packagemanager/PackageManager/View.hpp
```diff
@@ -34,27 +38,40 @@ class PluginSettingsView : public score::GlobalSettingsView
...
+  QPushButton* m_link{new QPushButton{tr("Open link")}};
   QPushButton* m_install{new QPushButton{tr("Install")}};
+  QPushButton* m_uninstall{new QPushButton{tr("Uninstall")}};
 
   QProgressBar* m_progress{new QProgressBar};
   QNetworkAccessManager mgr;
-
   int m_addonsToRetrieve = 0;
+
+  QStorageInfo storage{score::AppContext().settings<Library::Settings::Model>().getPackagesPath()};
```
jcelerier: maybe initialize that one in the .cpp instead in order to not leak the needed headers for that in the .hpp

jcelerier: (mainly for reducing compile times)

jcelerier (changes requested):

src/plugins/score-plugin-packagemanager/PackageManager/View.hpp
thibaudk: Thanks ! it's done

src/plugins/score-plugin-packagemanager/PackageManager/View.cpp
thibaudk: Strange, I cannot find this in my code
https://github.com/thibaudk/score/blob/wip/pkgMng/src/plugins/score-plugin-packagemanager/PackageManager/View.cpp#L406

jcelerier:
code looks good, I think we can merge it ! and then future features can come in later pr. good for you ?

thibaudk:
yes sure !

Issue #1332, opened 2021-12-10
[Tempo Process] Hide automation when input is connected


thibaudk:

thibaudk:
closed with b6e4fd9

PR #1353, opened 2022-01-28
address #1332

thibaudk:


thibaudk:
fixed a typo

src/plugins/score-plugin-scenario/Scenario/Document/Tempo/TempoView.hpp
```diff
@@ -56,8 +57,51 @@ class TempoPresenter final
       QObject* parent)
       : CurveProcessPresenter{style, layer, view, context, parent}
   {
+    // only disable the curve when using the first 2 inlets
+    for (int i = 0; i < 2; i++)
+    {
+      QObject::connect(
```
jcelerier, line 63: so, here I'd recommend just having a lambda instead of the two functions: 


      QObject::connect(layer.inlets()[i], &Process::Inlet::cablesChanged,
                                      this, [] { disableIfNeeded(); }); 
                                      
it should work for both cases and take less space.

src/plugins/score-plugin-scenario/Scenario/Document/Tempo/TempoView.hpp
```diff
@@ -56,8 +57,51 @@ class TempoPresenter final
       QObject* parent)
       : CurveProcessPresenter{style, layer, view, context, parent}
   {
+    // only disable the curve when using the first 2 inlets
+    for (int i = 0; i < 2; i++)
```
jcelerier, line 61: I think that you can do it for all the inlets (as right now there are three afaik ?)

src/plugins/score-plugin-scenario/Scenario/Document/Tempo/TempoView.hpp
```diff
@@ -56,8 +57,51 @@ class TempoPresenter final
...
+  }
+
+  void disableWithCables() { disableIfNeeded(); }
+
+  void disableIfNeeded()
+  {
+    bool should_disable{false};
+
+    // only disable the curve when using the first 2 inlets
+    for (int i = 0; i < 2; i++)
```
jcelerier, line 85: same than above

src/plugins/score-plugin-scenario/Scenario/Document/Tempo/TempoView.hpp
```diff
@@ -56,8 +57,51 @@ class TempoPresenter final
...
+    {
+      const auto inlet{m_process.inlets()[i]};
+
+      if (inlet->address().toString_unsafe() != ":/"
```
jcelerier: hmm, there should be a better way to do that check, I'll look around in score (and add a method "isValid" or something to State::Address if that's not the case)

src/plugins/score-plugin-scenario/Scenario/Document/Tempo/TempoView.hpp
thibaudk: The third inlet is position, witch should probably not affect the curve.

src/plugins/score-plugin-scenario/Scenario/Document/Tempo/TempoView.hpp
thibaudk: I can add it in AddressAccessor if you wish

src/plugins/score-plugin-scenario/Scenario/Document/Tempo/TempoView.hpp
thibaudk: Yes of course !
Although I require this in the capture ie:
```
          [this] { disableIfNeeded(); });

```

jcelerier (approved):

jcelerier:
:+1: thanks !

Issue #1361, opened 2022-02-08
Add an exit button on the start splash screen


jcelerier:

jcelerier:
thanks @RobinMontferme !

PR #1366, opened 2022-02-11
Fix/splash screen exit

RobinMontferme:
Fixed issue  #1361.
Credits to @aklevy for the icon

### DIFF
--- a/src/app/StartScreen.hpp
+++ b/src/app/StartScreen.hpp
@@ -210,6 +210,7 @@ public:
   void openFile(const QString& file) W_SIGNAL( openFile, file)
   void openFileDialog() W_SIGNAL( openFileDialog )
   void loadCrashedSession() W_SIGNAL( loadCrashedSession )
+  void exitApp() W_SIGNAL(exitApp)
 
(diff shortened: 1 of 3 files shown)


src/app/StartScreen.hpp
```diff
@@ -283,7 +284,14 @@ StartScreen::StartScreen(const QPointer<QRecentFilesMenu>& recentFiles, QWidget*
     label->setPixmaps(score::get_pixmap(":/icons/load_examples_off.png"),score::get_pixmap(":/icons/load_examples_on.png"));
     label->setOpenExternalLink(true);
     label->move(100,label_y);
-    label_y += 50;
+    label_y += 35;
+  }
+  { // Exit App
+      InteractiveLabel* label = new InteractiveLabel{titleFont,qApp->tr("Exit"), "", this};
```
jcelerier, line 290: purely stylistic thing: here is a really good place to use `auto`, as it just repeats the typename twice (we know that `new InteractiveLabel` is going to give us a `InteractiveLabel*`

jcelerier (approved):

Issue #950, opened 2019-05-12
[device explorer] add 'Reconnect all' option in contextual menu


avilleret:
and maybe allow to reconnect a selection of devices
currently when I select several devices (with CTRL+clic) "Reconnect" option is disabled

jcelerier:
seems to work ! @RobinMontferme maybe it would be worth having disconnect too ?

RobinMontferme:
why not I can get onto that as well

PR #1368, opened 2022-02-17
added a way to disconnect multiple devices at once

RobinMontferme:
Added a way to disconnect multiple devices at once like mentionned in issue #950


jcelerier (approved):
looks good thanks :)

Issue #727, opened 2018-04-26
MIDI: copy, paste, etc etc


jcelerier:

avilleret:
adding basic editing to Piano Roll would be great

- [x] create / move / delete note
- [x] change note pitch and velocity
- [ ] split pianoroll at given timecode / cursor

thibaudk:
Impossible to select Midi notes in Piano Roll.
"Lasso" selection is always required.

jcelerier:
selection should be fixed now

arthur-parmentier:
My own wishes for the piano roll (tested on win 10 Ossia 2.52):

- Changing Min and Max from the inspector does not update the number of rows and note labels in the background
- A visual keyboard would be more suitable than characters when a lot of keys are displayed on the vertical axis
- Getting visual clues by changing the pointer icon when the user is able to perform different actions would be nice. For instance, a specific icon should appear when hovering over the not onset/offset in order to modify them, while an other icon would be used to move the note...
- Suppr + ctrl-Z (windows) makes the piano roll reappear... without the note labels
- Suppr on selected midi note does nothing, whereas we would expect the note to be deleted

jcelerier:
all the infrastructure is in place, just need to fill these two functions: https://github.com/ossia/score/blob/master/src/plugins/score-plugin-midi/Midi/MidiNoteEditor.cpp

PR #1369, opened 2022-02-18
start copy and paste for midi notes

RobinMontferme:
Started to implement code to copy and paste Midi note in the Midi Piano Roll process from issue #727

### DIFF
--- a/src/plugins/score-plugin-midi/Midi/MidiNoteEditor.cpp
+++ b/src/plugins/score-plugin-midi/Midi/MidiNoteEditor.cpp
@@ -1,23 +1,109 @@
 #include "MidiNoteEditor.hpp"
 
+#include <Midi/Commands/AddNote.hpp>
 #include <Midi/Commands/RemoveNotes.hpp>
 #include <Midi/MidiNote.hpp>
+#include <Midi/MidiPresenter.hpp>
 #include <Midi/MidiProcess.hpp>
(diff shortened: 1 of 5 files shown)


src/plugins/score-plugin-midi/Midi/MidiNoteEditor.cpp
```diff
@@ -1,22 +1,98 @@
...
+
+  // TODO this is a bit lazy.. find a better positoning algorithm7
+  //Position in the case the pasting is done through ui
+  //  if (!mv.contains(mv_pt))
+  //    mv_pt = mv.mapToScene(mv.boundingRect().center());
+
+  // Read the copy json. TODO: give it a better mime type
+  //  auto origin = pres->toScenarioPoint(sv_pt);
+  auto obj = readJson(mime.data("text/plain"));
+  JSONWriter w(obj);
```
jcelerier, line 75: add an if(!obj.IsObject()) { return false; } here as this will crash otherwise

jcelerier (approved):
all good so far, just make sure to check the curve copy paste (CurveEditor.cpp) for some of the refactored parts

jcelerier:
code looks good so far ! I think that for the TODO of keepign the notes selected, you can do it in a further PR ?

RobinMontferme:
Sure absolutely !

Issue #1367, opened 2022-02-14
midi/osc learn autoscroll


gaeljaton:
minor enhancement, could be easier to check inputs when 'learn' if last input is visible (autoscroll new input when box is full).
kiss!

jcelerier:
@RobinMontferme  if you feel like it ? relevant class is LearnDialog in DeviceExplorerWidget.cpp

jcelerier:
thanks @RobinMontferme

PR #1375, opened 2022-03-10
made device learn autoscroll to the end when learning

RobinMontferme:
Fixed issue  #1367 
EDIT:
Also added autoscroll to the bottom on address learning in device view as proposed by @jcelerier   [here](https://github.com/ossia/score/pull/1375#issuecomment-1064941194)


jcelerier (commented):
perfect ! can you just change the commit message to fit the commit style and be a bit more explicit ? for this one 

[learn] Autoscroll learn dialog on new message

would be great

jcelerier:
also there is another learn dialog which is used when right clicking on an address, it would be cool if you could do the same to it. Check AddressLineEdit it should be around there

jcelerier (approved):
perfecto

PR #1380, opened 2022-03-31
[Fx] rename ValueFilter.hpp to Smooth.hpp

thibaudk:


jcelerier (approved):
yep sorry for not accepting this earlier

PR #1382, opened 2022-04-14
[addition] Virtual Midi Device Creation + midi devices bugfixes

RobinMontferme:
- Enabled creation of virtual Midi device from device explorer
- Fixed a bug where Devices weren't having their default settings applied
- Modified default settings for `MIDIInput` and `MIDIOutput` devices to set the correct port direction and added the new virtual parameter 
- Made so that the name of the device is also the name of the device in JACK
- made `MIDIInputProtocolFactory` and `MIDIOutputProtocolFactory` classes member public

### DIFF
--- a/src/plugins/score-plugin-protocols/Protocols/MIDI/MIDIProtocolSettingsWidget.cpp
+++ b/src/plugins/score-plugin-protocols/Protocols/MIDI/MIDIProtocolSettingsWidget.cpp
@@ -31,14 +31,23 @@ MIDIInputSettingsWidget::MIDIInputSettingsWidget(QWidget* parent)
   m_name = new State::AddressFragmentLineEdit{this};
   checkForChanges(m_name);
   m_createWhole = new QCheckBox{tr("Create whole tree"), this};
+  m_virtualPort = new QCheckBox{tr("Virtual Port"), this};
 
   auto lay = new QFormLayout;
   lay->addRow(tr("Name"), m_name);
   lay->addRow(m_createWhole);
+  lay->addRow(m_virtualPort);
 
   setLayout(lay);
   m_createWhole->setChecked(false);
-  m_createWhole->setEnabled(true);
+  m_createWhole->setEnabled(false);
+  m_virtualPort->setChecked(false);
+#if defined(_WIN32)
+  m_virtualPort->setEnabled(
+      false); //virtual port creation is unsupported by Windows
+#else
+  m_virtualPort->setEnabled(true);
+#endif
 }
 
 Device::DeviceSettings MIDIInputSettingsWidget::getSettings() const
@@ -49,7 +58,8 @@ Device::DeviceSettings MIDIInputSettingsWidget::getSettings() const
   s.name = m_name->text();
   s.protocol = MIDIInputProtocolFactory::static_concreteKey();
   midi.createWholeTree = m_createWhole->isChecked();
-
+  midi.virtualPort = m_virtualPort->isChecked();
+  midi.name = s.name;
   s.deviceSpecificSettings = QVariant::fromValue(midi);
 
   return s;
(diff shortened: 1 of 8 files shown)


src/plugins/score-plugin-protocols/Protocols/MIDI/MIDIProtocolSettingsWidget.cpp
```diff
@@ -31,14 +31,20 @@ MIDIInputSettingsWidget::MIDIInputSettingsWidget(QWidget* parent)
...
   lay->addRow(tr("Name"), m_name);
   lay->addRow(m_createWhole);
+  lay->addRow(m_virtualPort);
 
   setLayout(lay);
   m_createWhole->setChecked(false);
-  m_createWhole->setEnabled(true);
+  m_createWhole->setEnabled(false);
+  m_virtualPort->setChecked(false);
+#if !__WIN32
```
jcelerier: I don't think this works as "enabled" is the default state, it should be 

```
#if defined(_WIN32) // note the single underscore
  m_virtualPort->setEnabled(false);
#endif
```

instead (with maybe a comment to say that this is because win32 does not support that)

src/plugins/score-plugin-protocols/Protocols/MIDI/MIDIProtocolSettingsWidget.cpp
```diff
@@ -83,14 +93,20 @@ MIDIOutputSettingsWidget::MIDIOutputSettingsWidget(QWidget* parent)
...
   auto lay = new QFormLayout;
   lay->addRow(tr("Name"), m_name);
   lay->addRow(m_createWhole);
+  lay->addRow(m_virtualPort);
 
   setLayout(lay);
   m_createWhole->setChecked(false);
   m_createWhole->setEnabled(true);
+  m_virtualPort->setChecked(false);
+#if __WIN32
```
jcelerier: Same

src/plugins/score-plugin-protocols/Protocols/MIDI/MIDISpecificSettingsSerialization.cpp
```diff
@@ -44,4 +45,8 @@ void JSONWriter::write(Protocols::MIDISpecificSettings& n)
     n.api = (libremidi::API)it->toInt();
   else
     n.api = libremidi::default_platform_api();
+  if (auto it = obj.tryGet("VirtualPort"))
```
jcelerier, line 48: perfect thanks

src/plugins/score-plugin-protocols/Protocols/MIDI/MIDIProtocolSettingsWidget.cpp
```diff
@@ -120,13 +143,6 @@ void MIDIOutputSettingsWidget::setSettings(
     ossia::net::sanitize_device_name(prettyName);
   }
   m_name->setText(prettyName);
-
-  if (settings.deviceSpecificSettings.canConvert<MIDISpecificSettings>())
```
jcelerier, line 124: hmmm why was this part removed ?

jcelerier (changes requested):

smoothdeveloper:
🎉

src/plugins/score-plugin-protocols/Protocols/MIDI/MIDIProtocolSettingsWidget.cpp
RobinMontferme: Done

jcelerier (approved):
:+1:

jcelerier:
erf there's a conflict with libossia because of the merge of the other commit. Can you : 

- go into master
- git pull; git submodule update --init --recursive
- go into your branch
- git add 3rdparty/libossia # it should set it to the latest commit
- git commit --amend
- git push --force

and we'll be able to merge

PR #1395, opened 2022-04-25
[Modify] QPainterAdaptater

ErkEntonio:
Create some functions in QPainterAdaptater

### DIFF
--- a/src/plugins/score-plugin-avnd/Crousti/Painter.hpp
+++ b/src/plugins/score-plugin-avnd/Crousti/Painter.hpp
@@ -5,6 +5,11 @@
 #include <QGraphicsSceneMouseEvent>
 #include <halp/texture.hpp>
 
+#include <QGradient>
+#include <QPolygon>
+
+#include <cmath>
+
 namespace oscr
 {
 struct QPainterAdapter
@@ -82,6 +87,7 @@ struct QPainterAdapter
       p.setColor(qRgba(c.r, c.g, c.b, c.a));
       painter.setPen(p);
     }
+
     void set_stroke_width(double w)
     {
       QPen p = painter.pen();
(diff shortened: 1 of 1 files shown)


src/plugins/score-plugin-avnd/Crousti/Painter.hpp
```diff
@@ -151,6 +199,120 @@ struct QPainterAdapter
     {
         path.addEllipse(QPointF{cx, cy}, cr, cr);
     }
+
+    // Antoine
+    //          tab, count
+    void draw_polygon(double* tab, int count)
```
jcelerier, line 196: Yep

src/plugins/score-plugin-avnd/Crousti/Painter.hpp
```diff
@@ -151,6 +199,120 @@ struct QPainterAdapter
...
+            x = tab[i];
+            y = tab[i+1];
+            poly << QPointF(x, y);
+        }
+        path.addPolygon(poly);
+    }
+
+    // Antoine
+    //          x, y, radius1, radius2, count
+    void draw_star(float x, float y, float r1, float r2, int count)
```
jcelerier: This function should be expressed as a separate algorithm which uses this class I think

src/plugins/score-plugin-avnd/Crousti/Painter.hpp
```diff
@@ -151,6 +199,120 @@ struct QPainterAdapter
...
+        for(int i = 1; i < 2*count; i++){
+            path.lineTo(points[i]);
+        }
+
+        painter.drawPath(path);
+    }
+
+    //Antoine
+    //          w, h, ch, cm, cs
+    void draw_analog_clock(double w, double h, halp::rgba_color ch, halp::rgba_color cm, halp::rgba_color cs)
```
jcelerier: Same here

src/plugins/score-plugin-avnd/Crousti/Painter.hpp
```diff
@@ -94,6 +104,33 @@ struct QPainterAdapter
       painter.setBrush(QColor(qRgba(c.r, c.g, c.b, c.a)));
     }
 
+    // Antoine
+    //          x1, y1, x2, y2, c1, c2
+    void set_linear_gradient(double x1, double y1, double x2, double y2, halp::rgba_color c1, halp::rgba_color c2){
```
jcelerier, line 104: Those three functions are good :)

jcelerier:
So, for this commit what you should do is split it : 

- Keep the gradient methods and the draw_polygon method as they look like fundamental building blocks that we cannot express otherwise

- For the draw_star / draw_analog_clock, they aren't "fundamental units of drawing" and thus should instead be implemented "in terms of" the other functions of this class (which shouldn't be too hard). And for instance they could go in an "Example widgets" folder in avendish, not in the score codebase

jcelerier (approved):

PR #1399, opened 2022-05-10
[wip] Supporting more midi messages in MidiDrop plugin

smoothdeveloper:
This is wip of a first iteration on supporting a bit more of the midi messages in Score.

# current scope

* [ ] parsing
  * [x] cc
  * [ ] pitchbend
* [ ] ui
  * [ ] display enveloppes in the track
  * [ ] show notes with different colour per midi channel
* [ ] playback
  * [ ] output the additional events

Guidance appreciated, including pointers for:
* how to deal with the streaming of midi data on the output port of the track
* how to deal with the UI, to add a display of enveloppes for cc and pitchbend data

### DIFF
--- a/src/plugins/score-plugin-midi/Midi/MidiDrop.cpp
+++ b/src/plugins/score-plugin-midi/Midi/MidiDrop.cpp
@@ -69,7 +69,7 @@ void DropHandler::dropData(
               track.notes.apply_scale_ratio(ratio);
             }
             disp.submit(new Midi::ReplaceNotes{
-                          midi, track.notes, track.min, track.max, actualDuration});
+                          midi, track.notes, track.notes.minimum_pitch_noticed(), track.notes.maximum_pitch_noticed(), actualDuration});
           };
           vec.push_back(std::move(p));
         }
@@ -273,11 +273,7 @@ static void parseEvent_format0(const libremidi::track_event& ev, std::vector<Mid
         note.setStart(delta * (tick / total));
         note.setPitch(pitch);
         note.setVelocity(vel);
-        if (note.pitch() < nv.min)
-          nv.min = note.pitch();
-        else if (note.pitch() > nv.max)
-          nv.max = note.pitch();
-
+        nv.notes.notice_pitch(note.pitch());
         notes.insert({note.pitch(), note});
       }
       else
@@ -287,7 +283,7 @@ static void parseEvent_format0(const libremidi::track_event& ev, std::vector<Mid
         {
           NoteData note = it->second;
           note.setDuration(delta * (tick / total - note.start()));
-          nv.notes.push_back(note);
+          nv.notes.append(note);
         }
         notes.erase(pitch);
       }
@@ -305,7 +301,7 @@ static void parseEvent_format0(const libremidi::track_event& ev, std::vector<Mid
       {
         NoteData note = it->second;
         note.setDuration(delta * (tick / total - note.start()));
-        nv.notes.push_back(note);
+        nv.notes.append(note);
       }
       notes.erase(ev.m.bytes[1]);
       break;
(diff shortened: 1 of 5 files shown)


src/plugins/score-plugin-midi/Midi/MidiDrop.hpp
```diff
@@ -18,11 +18,101 @@ class DropHandler final : public Process::ProcessDropHandler
...
+  midi_size_t velocity;
+};
+struct NoteOffData {
+  midi_size_t channel;
+  midi_size_t note;
+  midi_size_t velocity;
+};
+struct MidiTrackEvent {
+  static MidiTrackEvent make_note_off(const double start, const midi_size_t ch, const midi_size_t n, const midi_size_t v){
+    return MidiTrackEvent{m_start: start, m_message:  Midi::NoteOffData{channel: ch, note: n, velocity: v}};
```
jcelerier: {m_start: start,

is not valid C++ syntax AFAIK, it should be 

     .m_start = start

 :)

src/plugins/score-plugin-midi/Midi/MidiDrop.hpp
```diff
@@ -18,11 +18,101 @@ class DropHandler final : public Process::ProcessDropHandler
       const score::DocumentContext& ctx) const noexcept override;
 };
 
+struct PitchbendData{
```
jcelerier, line 22: MAybe all of these could go into a more general Midi.hpp ... or, I wonder if it wouldn't be best in libremidi, and then try to consume those directly ? not sure...

src/plugins/score-plugin-midi/Midi/MidiDrop.cpp
```diff
@@ -65,11 +65,8 @@ void DropHandler::dropData(
             const double ratio = song_t / actualDuration.msec();
             if (ratio != 1.)
             {
-              for (auto& note : track.notes)
-              {
```
jcelerier: :+1:

src/plugins/score-plugin-midi/Midi/MidiDrop.hpp
```diff
@@ -18,11 +18,101 @@ class DropHandler final : public Process::ProcessDropHandler
...
+  }
+  double m_start{};
+
+  void setStart(const double start){ m_start = start; }
+  const double start(){return m_start; }
+  std::variant<Midi::NoteOnData, Midi::NoteOffData, Midi::ControllerData,Midi::PitchbendData> m_message;
+};
+
+struct MidiTrackEvents {
+  void push_back(double delta, int tick, double total, Midi::ControllerData c){
```
jcelerier, line 61: I wonder if "push_back" is a good name for something that doesn't behave like std::vector::push_back ; it looks closer to emplace_back (but I think that just "add" would be simpler)

src/plugins/score-plugin-midi/Midi/MidiDrop.hpp
```diff
@@ -18,11 +18,101 @@ class DropHandler final : public Process::ProcessDropHandler
...
+  midi_size_t note;
+  midi_size_t velocity;
+};
+struct NoteOffData {
+  midi_size_t channel;
+  midi_size_t note;
+  midi_size_t velocity;
+};
+struct MidiTrackEvent {
+  static MidiTrackEvent make_note_off(const double start, const midi_size_t ch, const midi_size_t n, const midi_size_t v){
```
jcelerier, line 47: these functions look a lot like the ones in libremidi, maybe it should go there ?

src/plugins/score-plugin-midi/Midi/MidiDrop.hpp
smoothdeveloper: image

Mmmh, ok, learning that I like GNU old-style (probably useful when C/C++ didn't agree yet on non positional saving lives in codebases terms) 🙂 

How can I make sure my PR isn't going to add warnings?

src/plugins/score-plugin-midi/Midi/MidiDrop.hpp
smoothdeveloper: Thanks for the suggestion, I'll start moving it soon then, please let me know what makes most sense.

I was going with MidiTrackEvent as type name (which is the core object along MidiNote which has "pointers" to optional? pair), and considering a "domain" folder:

/domain/MidiNote.hpp moving the current one
/domain/MidiTrackEvent.hpp the detailed header
/domain/Midi.hpp a logical header, including several others, for convenience in other modules.

Do you like this?

jcelerier:
hello ! just so you know - as this was requested in many places I added a very small separate "midi reader" object which will take a midi file explicitly and output every message, as well as a "midi filter" object which is able to filter a specific CC, pitch bend, after touch, etc... message. It's availble in the current nightlies

PR #1404, opened 2022-06-02
initial work on range slider

jcelerier:


src/lib/score/graphics/widgets/QGraphicsRangeSlider.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+QGraphicsRangeSlider::QGraphicsRangeSlider(QGraphicsItem* parent)
+{
+  auto& skin = score::Skin::instance();
+  setCursor(skin.CursorPointingHand);
+}
+
+void QGraphicsRangeSlider::setStart(double start)
+{
+  if (m_start != start)
```
jcelerier, line 20: in general, clamping has to occur before the check if there is one (it's not super necessary here), otherwise it can make loops

src/lib/score/graphics/widgets/QGraphicsRangeSlider.cpp
```diff
@@ -0,0 +1,138 @@
...
+  {
+    m_min = min;
+    m_max = max;
+    update();
+  }
+}
+
+void QGraphicsRangeSlider::mousePressEvent(QGraphicsSceneMouseEvent* event)
+{
+  if (event->button() == Qt::LeftButton)
```
jcelerier, line 72: add `using namespace std;` at the top of functions which use abs (or math functions in general)

src/lib/score/graphics/widgets/QGraphicsRangeSlider.cpp
```diff
@@ -0,0 +1,138 @@
...
+  {
+    d2s = abs(event->pos().x() - m_start * boundingRect().width());
```
jcelerier, line 74: boundingRect() can be expensive to call so it's better to call it once at the beginning: 


     const auto rect = boundingRect();
     d2s = abs(event->pos().x() - m_start * rect.width());
     .. etc ..

src/lib/score/graphics/widgets/QGraphicsRangeSlider.cpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+    else
+      handle = NONE;
+  }
+}
+
+void QGraphicsRangeSlider::mouseMoveEvent(QGraphicsSceneMouseEvent* event)
+{
+  switch (handle)
+  {
```
jcelerier, line 97: Same than above (and since it's only the width which is used, it 's even better and more readable to store it at the beginning

src/lib/score/graphics/widgets/QGraphicsRangeSlider.cpp
```diff
@@ -0,0 +1,138 @@
...
+    QWidget* widget)
+{
+
+  auto& skin = score::Skin::instance();
+  painter->setRenderHint(QPainter::Antialiasing, true);
+
+  painter->fillRect(boundingRect(), skin.Emphasis2.main.brush);
+  painter->fillRect(
+      rangeRect,
+      skin.Emphasis2.lighter180.brush); ///TODO: use right brush from skin here
```
jcelerier, line 142: it's `skin.Base4`

src/lib/score/graphics/widgets/QGraphicsRangeSlider.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+  explicit QGraphicsRangeSlider(QGraphicsItem* parent);
+
+  void setStart(double start);
+  void setEnd(double end);
+  void setRange(double min, double max);
+
+  void setValue(ossia::vec2f value);
+  ossia::vec2f value() const noexcept;
+  void setExecutionValue(ossia::vec2f) { } // TODO
```
jcelerier: it would be nice to have those: they show the current running value in the execution engine. Could be a small line at the bottom for instance

src/lib/score/graphics/widgets/QGraphicsRangeSlider.hpp
```diff
@@ -0,0 +1,74 @@
...
+  QRectF boundingRect() const override;
+  void paint(
+      QPainter* painter,
+      const QStyleOptionGraphicsItem* option,
+      QWidget* widget) override;
+  qreal y_factor{0.005};
+  qreal d2s, d2c, d2e, ypos, ydiff;
+  double val1, val2;
+
+  enum Handle
```
jcelerier, line 65: since this is only used in the implementation I'd put this enum in the cpp (and as `enum class` as those are pretty generic names)

src/lib/score/graphics/widgets/QGraphicsRangeSlider.hpp
```diff
@@ -0,0 +1,74 @@
...
+  {
+    START,
+    CENTER,
+    END,
+    NONE
+  };
+  Handle handle;
+
+  QRectF rangeRect{
```
jcelerier, line 74: I think this should be done in the constructor instead

src/plugins/score-lib-process/Process/Dataflow/ControlWidgets.hpp
```diff
@@ -288,6 +288,125 @@ struct IntSlider
...
+
+  template <typename T, typename Control_T>
+  static auto make_widget(
+      const T& slider,
+      Control_T& inlet,
+      const score::DocumentContext& ctx,
+      QWidget* parent,
+      QObject* context)
+  {
+    // TODO
```
jcelerier, line 306: it's not really urgent, maybe we can just show a label for now

jcelerier:
can I merge this @bltzr ?

bltzr:
I haven't checked your comments and suggestions, and it wasn't finished (values where not stored, e.g.) the last time I worked on it, but if you made it work, then yeah, sure

jcelerier:
comments fixed and PR merged

PR #1420, opened 2022-07-18
Fix source typos

luzpaz:


jcelerier (approved):
thanks ! i think the typos come right from stackoverflow here...

PR #1425, opened 2022-09-01
[tools] added an option to input build directory path

RobinMontferme:
Added a way to specify build directory path as an argument whern executing [developper.sh](tools/developer.sh).
- If left blank (default behavior) creates the build directory in the root dir of score
- If a path if passed, will create  a directory, configure the project and build it in.

### DIFF
--- a/tools/developer.sh
+++ b/tools/developer.sh
@@ -3,7 +3,6 @@ echo "Running on OSTYPE: '$OSTYPE'"
 DISTRO=""
 CODENAME=""
 SUDO=$(command -v sudo 2>/dev/null)
-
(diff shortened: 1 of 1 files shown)


jcelerier (changes requested):
Instead of duplicating the code three times, you could instead at the top of the file, after SCORE_PATH=$PWD, do : 

```bash
  if [[ "$#" -eq 0 ]];  then
    BUILD_DIR=build-developer
  else
    BUILD_DIR=$1
  fi
  ```

and then just do `mkdir -p $BUILD_DIR`

jcelerier (approved):
looks good, thanks !

Issue #477, opened 2017-06-01
oscquery device 'disconnected' state should be more visible


pchdev:
currently, the device's name is in italic when disconnected, which is not really visible in the device explorer. Maybe put it in red? or with some sort of highlight?

LouisHLAV:
Hey!
I'm interested in tackling this issue.
Would appending  "(disconnected)" to the device's italicized name help? E.g., _`Device X (disconnected)`_

![Example of connected vs. disconnected devices](https://user-images.githubusercontent.com/43889220/197309446-9963429f-5712-4fdf-99df-6f4ad2fff948.png)

jcelerier:
hello !
yes, that would be a pretty good way to do it I think :)

it should be relatively straightforward in the code: simply check for "disconnected" in the "name" case here and return the modifed version.

 https://github.com/ossia/score/blob/master/src/plugins/score-lib-device/Device/ItemModels/NodeDisplayMethods.cpp#L118 (function `deviceNameColumnData`)

please ping me directly here or on discord if you have issues with the build, I recommend using this script to set-up a proper development environment & build: https://github.com/ossia/score/blob/master/tools/developer.sh !

jcelerier:
closing, thanks again @LouisHLAV !

PR #1433, opened 2022-10-25
[ui] Make device disconnected state more visible

LouisHLAV:
This patch is a proposed fix for #477: the `(disconnected)` string is appended to a device's name in the Device Explorer if it's disconnected.

This is my first contribution the Score project - and one of my first to OSS projects -, so I apologize in advance if I'm doing something wrong haha

### DIFF
--- a/src/plugins/score-lib-device/Device/ItemModels/NodeDisplayMethods.cpp
+++ b/src/plugins/score-lib-device/Device/ItemModels/NodeDisplayMethods.cpp
@@ -127,11 +127,12 @@ QVariant deviceNameColumnData(const Device::Node& node, bool connected, int role
 
   switch(role)
   {
-    case Qt::DisplayRole:
-    case Qt::EditRole: {
+    case Qt::DisplayRole: {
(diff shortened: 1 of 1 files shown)


jcelerier:
thanks a lot and welcome to the team :) ! 

the patch just needs a little bit more work :

Here it will add (disconnected) to both the DisplayRole and EditRole.
DisplayRole is what you want: this is the text that is being shown when just looking at the tree.

EditRole is not entirely yet implemented in score but the code here has to be correct for when it will be: what is returned by EditRole is the string that is being shown to the user when they double-click in a field of a table / tree so that they can't change the name of the device by typing it out directly (again this specific part is not implemented yet but we just don't want to have a hidden bug lurking there for the day we implement it :p)

LouisHLAV:
Thanks for the review and for the warm welcome, @jcelerier!
That makes total sense. I was wondering why the EditRole branch was there XD
I'll upload a fixed patch shortly :)

jcelerier (approved):
:+1: perfect thanks!

PR #1434, opened 2022-11-02
find'n'replace

jcelerier:


src/plugins/score-lib-state/State/Expression.cpp
```diff
@@ -27,6 +27,65 @@ bool operator<(const State::ExprData& lhs, const State::ExprData& rhs)
 {
   return lhs.toString() < rhs.toString();
 }
+
+bool findAddressInExpression(const State::Expression& expr, const State::Address& addr){
+
```
jcelerier, line 32: clang-format :D
if you want you can just enable it so that it auto formats on save in qt creator

src/plugins/score-lib-state/State/Expression.hpp
```diff
@@ -32,7 +32,7 @@ enum class UnaryOperator
 
 struct SCORE_LIB_STATE_EXPORT ExprData
     : public score::VariantBasedNode<Relation, Pulse, BinaryOperator, UnaryOperator>
-{
+  {
```
jcelerier, line 35: wop?

src/plugins/score-plugin-scenario/Scenario/Inspector/ObjectTree/SearchReplaceWidget.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+void SearchReplaceWidget::replace()
+{
+  auto& ctx = m_ctx.documents.currentDocument()->context();
+  MacroCommandDispatcher<Command::ReplaceAddresses> disp{ctx.commandStack};
+  if(!m_newAddress.isSet())
+  {
+    return;
+  }
+  else
```
jcelerier: in this case (`if(xx) return;`) don't put a else, just put it after, this way there is no unneeded indentation level. 
this even has a name: return-early pattern https://medium.com/swlh/return-early-pattern-3d18a41bba8  (and it's a good one, i try to use it as much as possible)

src/plugins/score-plugin-scenario/Scenario/Inspector/ObjectTree/SearchReplaceWidget.cpp
```diff
@@ -0,0 +1,327 @@
...
+  {
+    for(auto* obj : m_matches)
+    {
+      if(auto state = qobject_cast<const StateModel*>(obj))
+      {
+        auto& root = state->messages().rootNode();
+        //TODO maybe implement a better ver of RenameAddressInState ?
+        disp.submit(new Command::RenameAddressInState(
+            *state, State::AddressAccessor{m_oldAddress},
+            State::AddressAccessorHead{m_newAddress.path.back(), {}}));
```
jcelerier: bug here: if the address is like "foo:/" then path will be empty and this will crash.
for now you can just add a `if(!m_newAddress.path.isEmpty())` guard in the beginning

src/plugins/score-plugin-scenario/Scenario/Inspector/ObjectTree/SearchReplaceWidget.cpp
```diff
@@ -0,0 +1,327 @@
...
+      else if(auto ts = qobject_cast<TimeSyncModel*>(obj))
+      {
+        auto expr = ts->expression();
+        State::replaceAddress(expr, m_oldAddress, m_newAddress);
+        disp.submit(new Command::SetTrigger(*ts, expr));
+      }
+
+      else if(auto inlet = qobject_cast<Process::Inlet*>(obj))
+      {
+        disp.submit(new Process::ChangePortAddress(*inlet, m_newAddress));
```
jcelerier: I wonder if you can maybe just cast to Process::Port to only have one branch here? instead of two for inlet / outlet

jcelerier (commented):
small things but looks mostly good !

jcelerier:
merged it but there are a few remaining issues, let's do them in follow up commits

PR #1447, opened 2022-12-28
[avendish] frist attempt at making devices with avendish addons

thibaudk:
!!! for reviews only at the moment !!!

I am trying to make a generic device with a avendish "Node" as a template argument. This may be a complete wrong way to go about it.
for example, I am trying to provide the name and uuid from ```avnd::get_uuid<Node>()``` and ```avnd::get_name<Node>()```


thibaudk:
Done by @jcelerier in https://github.com/ossia/score/commit/18d736281626f0e25ac81338697de72b627ba68c

PR #1450, opened 2023-01-11
Fix #1440: save the correct name of scenarios in preset files

### LINKED ISSUE #1440: name of node when saved as preset
When you drag a node into the user library to save it as a preset, it always gets named "Scenario.scenario" It would make sense if it gets the name of the node.

yiids:


jcelerier (approved):
looks good! thanks :)

PR #1497, opened 2023-12-17
fix GenCodeOptLevel vs GenCodeOpt

dave2309:
ref #1496


dave2309:
clang 18
debian sid

jcelerier:
Can you ifdef it with LLVM_VERSION? As score still needs to build against clang 15,16,17..
Thanks!

dave2309:
done

jcelerier (approved):
thanks!

PR #1542, opened 2024-06-17
Add 'NSBluetoothAlwaysUsageDescription' to info.plist to fix BLE on Mac

coral:
Ossia Score crashes on OSX when adding the BLE input because the OS terminates you for not providing a description on the Info.plist

Screenshot 2024-06-17 at 09 27 45

This adds this to the templated info.plist that's used for the Mac build.


jcelerier:
Thanks a lot ! this didn't come up in dev builds.. ah well

PR #1556, opened 2024-07-31
fix shmdata input do not reconnect when writer quits

nicobou:
This fix hot plug with Shmdata Input, when the writer quits and restart.

Without this fix, the Shmdata Input work when score is playing, but when the GStreamer pipeline is restarted, it requires the score timeline to be stopped and played again.

With this fix, the restart of the GStreamer pipeline maintain score's Shmdata Input updated with new video frames.


jcelerier:
ah, thanks for the fix ! I should have read the doc more :)

PR #1568, opened 2024-08-22
Processes documentaton links

HugoGenillier:
Description links were added for processes defined with process metadata.
(Avendish and node processes aren't there yet)


jcelerier:
thank you so much for this work !

samamou:
Looking for a global stylesheet where i could **set the color of external links** (to match the icon color) Is this possible? @jcelerier 

Or i can create a **dedicated stylesheet for QUrls** somewhere, somehow?

image

PR #1594, opened 2024-11-19
[isf] WIP on having a compute shader spec

jcelerier:


jcelerier:
Done for quite some time now

PR #1601, opened 2024-12-01
perf: prereserve some audio buffers

jcelerier:


jcelerier:
Closing: superseded upstream.

\`DocumentPlugin::makeGraph()\` already prereserves the audio buffer pool through the public API — 500 buffers at \`execState->bufferSize\`, versus the ≥100 at \`bs + 16\` proposed here — and \`reload()\`, the function this PR patches, calls \`makeGraph()\` first. The block added here would be a weaker duplicate running immediately after.

It also no longer compiles: libossia has since made \`audio_buffer_pool\` inherit \`object_pool<audio_channel>\` privately, exposing only \`acquire()\` / \`release()\`, so reaching into \`.buffers\` is rejected on every compiler (verified on gcc 15/16, clang 19/22).

PR #1604, opened 2024-12-01
core: Implement a command transaction system

jcelerier:


jcelerier:
A concrete use case for this, plus the two small pieces that seem to be missing for it to apply to preset loading.

## Symptom

Switching presets on an avnd audio process produces a burst of very loud notes that settles to the correct level shortly after.

## Path

Preset application converges on one command from four entry points:

| entry | site |
| --- | --- |
| preset menu | `EffectLayer.cpp:296` |
| preset cycling (next/prev) | `EffectLayer.cpp:412` |
| library drop → new process | `CommandAPI.cpp:237` (inside a macro, + `ChangeElementName`) |
| drop on cable / before / after port | `ProcessCreation.cpp` |

All go `LoadPresetCommandFactoryList::make` → `CommandDispatcher::submit` → `SendStrategy::send` → `CommandStack::redoAndPush`, whose redo reaches `Process::loadFixedControls`. That loops over the preset's entries calling `ControlInlet::setValue` one port at a time; each `valueChanged` lands in `Gfx::con_unvalidated::operator()`, which does

```cpp
ctx.executionQueue.enqueue(control_updater{node->controls[i], val});
```

— **one queue item per control**. For a process with a few hundred controls the UI thread is filling that queue for a while, and the audio thread drains it concurrently, so a tick can observe a partially applied preset.

## Why a partial application is loud rather than merely wrong

Gain is usually spread over more than one control. In the process I hit this with, each voice's output gain is a per-channel `Level`, a per-channel `Drive` and a global `Volume`. Computing, over every pair of shipped presets, the worst instantaneous gain reachable when each control may independently be old or new:

```
worst-case transient overshoot: 45.7 dB
  level 2.00 -> 0.07 | drive 12.6 -> 0.0 | master 0.50 -> 0.30
```

Any process whose gain staging is split across controls has the same exposure.

To be explicit about what is and isn't established: I confirmed offline (re-applying a full parameter set between blocks, with DSP state carried over) that carried-over node state does **not** cause this, and the numbers above are measured; that a tick actually lands mid-loop is inferred from the per-control enqueue rather than directly observed. Batching removes the possibility either way.

## What this PR already provides

`Execution::Transaction` on master is exactly the right primitive — `run_all()` enqueues a *single* queue item that applies every batched command, so the audio thread sees all-or-nothing. This PR adds the missing plumbing around it: the transaction pointer on `Execution::Context`, `execCommand()` routing into the active transaction, the `CommandTransaction` RAII counter, and the `DocumentPlugin` wiring.

## The two gaps

**1. Control updates bypass it.** `con_unvalidated::operator()` enqueues directly, so nothing from a preset load is captured:

```diff
- ctx.executionQueue.enqueue(control_updater{node->controls[i], val});
+ ctx.execCommand(control_updater{node->controls[i], val});
```

This is the single chokepoint for control→engine delivery for the avnd and gfx nodes, so it looks like the only site that needs changing.

**2. Nothing opens a transaction around a preset load.** As I read the diff, `undoQuiet`/`redoQuiet` are wrapped but `redoAndPush`/`push` are opt-in. Extending the same `CommandTransaction t{*this}` to `redoAndPush`, `push`, `undo` and `redo` would make a command the transaction boundary, which seems like the right unit — if one user action moves N controls they should reach the engine together. That covers preset load, preset cycling, undo of either, script reload and macro-driven multi-parameter edits with no per-feature code, and the nesting counter already makes the macro case (`CommandAPI` submits load-preset *and* `ChangeElementName`) safe.

The narrower alternative is to open a transaction at the four preset call sites only, but that is four edits instead of one and leaves the same hazard everywhere else.

Risk of the broader version looks low: it would only change behaviour for a command relying on one of its own execution commands taking effect partway through its own redo, which nothing can do today since the queue is drained asynchronously on the audio thread.

## Two adjacent things noticed while tracing this

- `loadFixedControls` does `ossia::find_if` over all inlets for **every** entry — O(N²), around 100k comparisons for a 330-control preset. Not a correctness issue, but it directly lengthens the window this PR would close.
- `ControlInlet::setValue` does not clamp to the port's domain, and presets carry no range information. A preset written against a build where a `spinbox_i32` had a larger maximum will happily push the old value into the new build; if the DSP uses that control as a loop bound over a fixed-size array, the result is an out-of-bounds write rather than a clamped parameter. Orthogonal to atomicity, but it is the other way a preset can misbehave across builds.

PR #1616, opened 2024-12-11
artnet: initial work towards a led strip ui

jcelerier:


jcelerier:
done better in https://github.com/ossia/score/pull/1706

PR #1619, opened 2024-12-11
Fix: Include QUrl

aalex:
Fixes a build error on Pop OS 22.04.

### DIFF
--- a/src/plugins/score-plugin-protocols/Protocols/Evdev/EvdevProtocolFactory.cpp
+++ b/src/plugins/score-plugin-protocols/Protocols/Evdev/EvdevProtocolFactory.cpp
@@ -17,6 +17,7 @@
 #include <QFormLayout>
 #include <QObject>
 #include <QRegularExpression>
+#include <QUrl>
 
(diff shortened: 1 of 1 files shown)


aalex:
I think there was another branch with the fix: https://github.com/ossia/score/pull/1613
And these changes might have been integrated into the main branch in another PR, anyways.

jcelerier:
Hm I don't see it in master, let's put it there and rebase wip/pico-next on master afterwards

jcelerier:
Argh, though this brings all the wip/pico-next commits with that... could you instead change this branch so that it is this commit 

https://github.com/ossia/score/pull/1619/commits/e58e54e74260a7216ec06be33870b7adfa3e0806

cherry-picked on top of current master ? 

(or just with the branch force-pushed as I think it would be equally fast to just redo the commit than to type git cherry-pick...)

aalex:
It should be better, now.

jcelerier:
thanks !

jcelerier (approved):

PR #1640, opened 2024-12-26
MIDI Remote control

jcelerier:
- **remotecontrol: move the current implementation in its own folder**
- **remotecontrol: put ws impl in ws namespace**
- **remotecontrol: implement mcu protocol**

TODO: fix support for VST2 plug-ins which use a different type for controls
=> done


src/plugins/score-lib-process/RemoteControl/RemoteControlProvider.hpp
jcelerier, line 1: @thibaudk likely somewhere the button grid implementation would either have a similar interface, or we'd extend this one so that it can cover both cases

thibaudk: the strategy for handling common buttons like "play", "pause", "up", "down"... in [bugui](https://codeberg.org/bugui/bugui/src/branch/master/include/midi_hardware.hpp#L214) relies on generic functions defined in the controller (software side) and the hardware. We check if an "on_play(bool)" function is defined on the software side, and compiles it only if a "play_id()" exists in the definition of the hardware device.

thibaudk: "on_play" could potentially be defined with different types as argument, allowing different behavior, depending on the specification of the hardware. This would be a generic differentiation between  "ControllerAction".
for example, "on_play(bool)" would be executed when a "on/off" play button is pressed. "on_play(float)" would be executed if the same button a different piece of hardware was pressure sensitive.

jcelerier: okay, so here I think the goal would be to define which things are available on the software side. Right now I'm working on this by extending https://github.com/ossia/score/blob/master/src/plugins/score-plugin-js/JS/Qml/EditContext.hpp - do you see any feature that I could add here that would be useful to you ?

jcelerier:
@thibaudk so on my side this looks like it works, are you ok with me merging it ?

PR #1647, opened 2024-12-27
Open code in external editor

### LINKED ISSUE #889: add "open in external editor" button to embedded editor
when clicking the button, Qt should :

1. create a temporary file with the content of the embedded editor
2. open external editor defined in the Score's preference panel with this temp file
3. watch file modification and update embedded editor accordingly (should this work in both way ?)

file watching example https://github.com/OSSIA/score/blob/master/base/plugins/score-plugin-js/JS/JSProcessModel.cpp#L78

samamou:
- [ ] Reuse previous editor instance instead of launching a new one each time and ending up with 30 sublime / vscode files open for each edit :')
   - [ ] before creating a new file, check if it exists?

### DIFF
--- a/src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
+++ b/src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
@@ -2,7 +2,8 @@
 
 #include "MultiScriptEditor.hpp"
 #include "ScriptWidget.hpp"
-#include "score/tools/FileWatch.hpp"
+
+#include <score/tools/FileWatch.hpp>
 
 #include <QCodeEditor>
 #include <QCoreApplication>
@@ -230,6 +231,7 @@ void ScriptDialog::stopWatchingFile(const QString& tempFile)
 
   auto& w = score::FileWatch::instance();
   w.remove(tempFile, m_fileHandle);
+  m_fileHandle.reset();
 }
 void MultiScriptDialog::openInExternalEditor(const QString& editorPath)
 {
--- a/src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
+++ b/src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
@@ -2,8 +2,10 @@
 
 #include "MultiScriptEditor.hpp"
 #include "ScriptWidget.hpp"
+#include "score/tools/FileWatch.hpp"
 
 #include <QCodeEditor>
+#include <QCoreApplication>
 #include <QDialogButtonBox>
 #include <QDir>
 #include <QFile>
@@ -176,16 +178,8 @@ void ScriptDialog::openInExternalEditor()
     return;
   }
 
-  if(!QFile::exists(editorPath))
-  {
-    QMessageBox::warning(
-        this, tr("Error"), tr("the configured external editor does not exist."));
-    return;
-  }
-
-  QString tempDir = QStandardPaths::writableLocation(QStandardPaths::TempLocation);
-  QString tempFile = tempDir + "/ossia_script_temp.js";
-
+  QString tempFile = QStandardPaths::writableLocation(QStandardPaths::TempLocation)
+                     + "/ossia_script_temp.js";
   QFile file(tempFile);
   if(!file.open(QIODevice::WriteOnly))
   {
@@ -196,12 +190,40 @@ void ScriptDialog::openInExternalEditor()
   file.write(this->text().toUtf8());
   file.close();
 
+  auto& w = score::FileWatch::instance();
+  m_fileHandle = std::make_shared<std::function<void()>>([this, tempFile]() {
+    QFile file(tempFile);
+    if(file.open(QIODevice::ReadOnly))
+    {
+      QString updatedContent = QString::fromUtf8(file.readAll());
+      file.close();
+
+      QMetaObject::invokeMethod(m_textedit, [this, updatedContent]() {
+        if(m_textedit)
+        {
+          m_textedit->setPlainText(updatedContent);
+        }
+      });
+    }
+  });
+  w.add(tempFile, m_fileHandle);
+
   if(!QProcess::startDetached(editorPath, QStringList() << tempFile))
   {
-    QMessageBox::warning(this, tr("Error"), tr("Failed to launch external editor"));
+    QMessageBox::warning(this, tr("Error"), tr("failed to launch external editor"));
   }
 }
 
+void ScriptDialog::stopWatchingFile(const QString& tempFile)
+{
+  if(tempFile.isEmpty())
+  {
+    return;
+  }
+
+  auto& w = score::FileWatch::instance();
+  w.remove(tempFile, m_fileHandle);
+}
 void MultiScriptDialog::openInExternalEditor()
 {
   QString editorPath = QSettings{}.value("Skin/DefaultEditor").toString();
--- a/src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
+++ b/src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
@@ -4,6 +4,7 @@
 #include "ScriptWidget.hpp"
 
 #include <score/tools/FileWatch.hpp>
+#include <score/widgets/SetIcons.hpp>
 
 #include <QCodeEditor>
 #include <QCoreApplication>
@@ -44,6 +45,22 @@ ScriptDialog::ScriptDialog(
   lay->setStretch(1, 1);
   auto bbox = new QDialogButtonBox{
       QDialogButtonBox::Ok | QDialogButtonBox::Reset | QDialogButtonBox::Close, this};
+  if(auto editorPath = QSettings{}.value("Skin/DefaultEditor").toString();
+     !editorPath.isEmpty())
+  {
+    auto openExternalBtn = new QPushButton{tr("Edit in default editor"), this};
+    connect(openExternalBtn, &QPushButton::clicked, this, [this, editorPath] {
+      openInExternalEditor(editorPath);
+    });
+    bbox->addButton(openExternalBtn, QDialogButtonBox::HelpRole);
+    openExternalBtn->setToolTip(
+        tr("Edit in the default editor set in Score Settings > User Interface"));
+    auto icon = makeIcons(
+        QStringLiteral(":/icons/undock_on.png"),
+        QStringLiteral(":/icons/undock_off.png"),
+        QStringLiteral(":/icons/undock_off.png"));
+    openExternalBtn->setIcon(icon);
+  }
   bbox->button(QDialogButtonBox::Ok)->setText(tr("Compile"));
   bbox->button(QDialogButtonBox::Reset)->setText(tr("Clear log"));
   connect(bbox->button(QDialogButtonBox::Reset), &QPushButton::clicked, this, [this] {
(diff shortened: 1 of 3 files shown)


src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +158,30 @@ void MultiScriptDialog::clearError()
   m_error->clear();
 }
 
+void ScriptDialog::openInExternalEditor()
+{
+  QString editorPath
```
jcelerier:
```cpp
editorPath= QSettings{}.value("Skin/DefaultEditor").toString();
```

samamou: yess 🙏

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +159,42 @@ void MultiScriptDialog::clearError()
...
+  QString tempFile = tempDir + "/ossia_script_temp.js";
+
+  QFile file(tempFile);
+  if(!file.open(QIODevice::WriteOnly))
+  {
+    QMessageBox::warning(this, tr("Error"), tr("failed to create temporary file."));
+    return;
+  }
+
+  QTextStream stream(&file);
```
jcelerier: I think you can just do 

     file.write(this->text().toUtf8());

samamou:
[QFileSystemWatcher](https://doc.qt.io/qt-6/qfilesystemwatcher.html) to monitor if a temp file changes externally + syncs updates back into the score editor? @jcelerier

jcelerier:
yep ! but keep in mind this issue: https://stackoverflow.com/questions/18300376/qt-qfilesystemwatcher-signal-filechanged-gets-emited-only-once/18304496 

so a bit of intelligence is needed - I think the best way is the answer that proposes watching the parent folder for newly created file, and check if it matches.

Also I think this behaviour should be encapsulated in a entirely separate utility class as there is this requirement in multiple other places in ossia :)

samamou:
oki another crash on thread 20 here [score-crash.txt](https://github.com/user-attachments/files/18324366/score-crash.txt)
trying to retrace what i did that caused this

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +167,114 @@ void MultiScriptDialog::clearError()
   m_error->clear();
 }
 
+void ScriptDialog::openInExternalEditor()
+{
+  QString editorPath = QSettings{}.value("Skin/DefaultEditor").toString();
+
+  if(editorPath.isEmpty())
```
jcelerier, line 198: Maybe we could just not show the widget if there's nothing in the settings ?

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +167,114 @@ void MultiScriptDialog::clearError()
...
+  {
+    QMessageBox::warning(
+        this, tr("Error"), tr("no 'Default editor' configured in score settings"));
+    return;
+  }
+
+  QString tempFile = QStandardPaths::writableLocation(QStandardPaths::TempLocation)
```
jcelerier: usually we'd mark something like this `const`

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +167,114 @@ void MultiScriptDialog::clearError()
...
+                     + "/ossia_script_temp.js";
+  QFile file(tempFile);
+  if(!file.open(QIODevice::WriteOnly))
+  {
+    QMessageBox::warning(this, tr("Error"), tr("failed to create temporary file."));
+    return;
+  }
+
+  file.write(this->text().toUtf8());
+  file.close();
```
jcelerier: it's not necessary to close, QFile's destructor does it

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +167,114 @@ void MultiScriptDialog::clearError()
...
+        if(m_textedit)
+        {
+          m_textedit->setPlainText(updatedContent);
+        }
+      });
+    }
+  });
+  w.add(tempFile, m_fileHandle);
+
+  if(!QProcess::startDetached(editorPath, QStringList() << tempFile))
```
jcelerier: in 2024 I'd recommend `{ tempFile }` instead of `QStringList() << tempFile` as it is closer to how it is done in most "non-Qt" C++

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +167,114 @@ void MultiScriptDialog::clearError()
...
+
+void ScriptDialog::stopWatchingFile(const QString& tempFile)
+{
+  if(tempFile.isEmpty())
+  {
+    return;
+  }
+
+  auto& w = score::FileWatch::instance();
+  w.remove(tempFile, m_fileHandle);
```
jcelerier, line 247: here I would also clear m_fileHandle: `m_fileHandle.reset()` to not keep its memory around

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -150,4 +167,114 @@ void MultiScriptDialog::clearError()
...
+}
+void MultiScriptDialog::openInExternalEditor()
+{
+  QString editorPath = QSettings{}.value("Skin/DefaultEditor").toString();
```
jcelerier: const QString

jcelerier (commented):
great, I have small nitpicks but the code looks very clean, !

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
```diff
@@ -2,12 +2,19 @@
 
 #include "MultiScriptEditor.hpp"
 #include "ScriptWidget.hpp"
+#include "score/tools/FileWatch.hpp"
```
jcelerier: I try to always keep the < > include style instead of " " in ossia as much as possible, as it is less ambiguous

src/plugins/score-lib-process/Process/Script/ScriptEditor.cpp
samamou: Yes, small scope question 

In order to have the button show up ONLY if the path is set in settings.. I need to declare editorPath in the MultiScriptDialog and ScriptDialog constructors

Is it better to declare it this way 

```
QString editorPath = QSettings{}.value("Skin/DefaultEditor").toString();

  if(!editorPath.isEmpty())
  {
    auto openExternalBtn = new QPushButton{tr("Open in external editor.."), this};
    connect(openExternalBtn, &QPushButton::clicked, this, [this, editorPath] {
      openInExternalEditor(editorPath);
    });
    lay->addWidget(openExternalBtn);
  }
```

or have a global member var 

```
m_editorPath = QSettings{}.value("Skin/DefaultEditor").toString();
if (!m_editorPath.isEmpty()) {
  auto openExternalBtn = new QPushButton{tr("Open in external editor.."), this};
  connect(openExternalBtn, &QPushButton::clicked, this, [this] {
    openInExternalEditor(m_editorPath);
  });
  lay->addWidget(openExternalBtn);
}
```

The differenece is also in the way the lambda captures the variable

jcelerier: so, the "general answer" to this question is "it depends".

The ingredients of this "it depends" in general are: 

- How many methods of the class are going to use the editorPath
- How heavy is editorPath to copy
- How strictly we want to adhere to the "variable scopes shall be as small as possible" mantra

For instance, every time we have a lambda that captures `editorPath` this does a QString copy (which is cheap as QString is a reference-counted, copy-on-write type - e.g. it does not copy the internal data until it is accessed through a non-const method). 

If we had like 4 lambdas that need it then it'd definitely be better to have it as a class member. Here I would say it is cleaner to keep it as a local variable since it's only needed once.

Ideal code to me looks like: 

```cpp
  if(auto editorPath = QSettings{}.value("Skin/DefaultEditor").toString(); 
    !editorPath.isEmpty())
  {
    auto openExternalBtn = new QPushButton{tr("Open in external editor.."), this};
    connect(openExternalBtn, &QPushButton::clicked, this, [this, editorPath] {
      openInExternalEditor(editorPath);
    });
    lay->addWidget(openExternalBtn);
  }
```

samamou:
I spoke to Lou and Zach briefly about scripting in other software to discuss the behavior they expect 

### Feedback: 

- **Workflow / sync** : they expect the external editor changes to update in score **automatically onSave()**, aligning with live-coding workflows  


- **Error feedback** : they both expect visual feedback indicating an issue (syntax error)  directly on or near the node in the scenario editor. Feedback from bespoke editor:
 

Possible quick implementation:
Screenshot 2025-01-13 at 2 50 30 PM 
 
It also got me thinking, is it possible to have this open up and direct the log/output here / to another terminal  instead of the QTextEditor + give feedback onCompile()/onRun() such as adding a comment  block *****SUCCESSFULLY COMPILED***** 

Screenshot 2025-01-13 at 2 50 30 PM

-  **Code MiniView (Internal viewer)** : they appreciate having a quick inline read-only view of the code without needing to open a full external editor. They said "Currently in score, a Dialog **blocks** your view of the Scenario Editor"  blocks feels like a heavy word, and i guess there could be a lighter way of presenting this information, for quick checking. For example, a collapsible inline "peek" panel? (not sure how this would like) 

[Nodes.io](https://nodes.io/)


[Max 9](https://cycling74.com/products/max-9)


vscode peek window



@jcelerier can you please tell me which is high/ low effort?

jcelerier:
Amazing thanks for this analysis !

jcelerier:
>     Workflow / sync : they expect the external editor changes to update in score automatically onSave(), aligning with live-coding workflows

sounds like what we discussed earlier, this is also what i'd expect as behaviour

> Error feedback : they both expect visual feedback indicating an issue (syntax error) directly on or near the node in the scenario editor. 

yep, that sounds doable (in a further PR!). Should be done here: https://github.com/jcelerier/QCodeEditor ; I also like the "red" error state that you propose, imho this one should be generalised as there are definitely non-code-editor things which can be in erroneous states like this!

For the immediate needs though, I think a bespoke-like thing with the line being underlined is the best bet. (again, in a further PR, this is another feature on top of this one.)


>  Code MiniView (Internal viewer) : they appreciate having a quick inline read-only view of the code without needing to open a full external editor. They said "Currently in score, a Dialog blocks your view of the Scenario Editor" blocks feels like a heavy word, and i guess there could be a lighter way of presenting this information, for quick checking. For example, a collapsible inline "peek" panel? (not sure how this would like)

yeah, this one's more of a challenge and to be thought of as part of a more general "debug overlays" feature, e.g. to see for instance not only code but also inputs & outputs signals, textures, etc.

src/plugins/score-lib-process/Process/Script/ScriptEditor.hpp
jcelerier, line 1: we need to do the same change in the MultiScriptEditor I think

samamou: yes i'm on it

samamou: Although... it works properly without adding to `MultiScriptEditor.hpp`

samamou: Ok... seems like it works without overriding actually, will remove from both..

jcelerier (approved):
LGTM! thanks!

PR #1685, opened 2025-01-19
feature/fulldome

jcelerier:
- **isf: set default range to 0-1 to not break mappings if the shader does not provide a range**
- **gfx: remove old Qt 5 Rhi compat headers**
- **libav: use undeprecated features**
- **3d: implement a fulldome projection mode suitable for point clouds and detailed geometry**


jcelerier:
@cbicari next time we can go play in the dome that's going to be funnnnn as now we can have "correct-ish" perspective in many cases for 3D content

PR #1687, opened 2025-01-20
Minimise build on snmalloc.

mjp41:
The `EXCLUDE_FROM_ALL` target on `add_subdirectory` means that it will only build items that are required to build other targets.  This means that it will not build the `memcpy` version that you are not using in this project, hence remove the issue with FORTIFY_SOURCE. This should also mildly improve your build times.


jcelerier:
Thanks !

PR #1713, opened 2025-02-08
feature/isf input for model display

jcelerier:
- **gfx: various small fixes, better handle when a shader does not have a material ubo**
- **gfx: tentative of allowing model display to directly use the parent shader if any**


jcelerier:
not necessary anymore with the Render Pipeline object

PR #1735, opened 2025-02-26
add multi-cursor manager plugin

D3mh4:


jcelerier:
hello! i think we can close this PR since it went into the GBAP repo, right ?

PR #1745, opened 2025-03-22
video: try a different tradeoff of video playback options

jcelerier:


jcelerier:
https://stackoverflow.com/questions/55186822/whats-ffmpeg-doing-with-avcodec-send-packet

PR #1754, opened 2025-04-03
isf preview: example of how making it go through vk

jcelerier:


jcelerier:
closed, next PR will use the chosen graphics api

PR #1768, opened 2025-04-17
fix/adjust_grid_size_to_generate_values_from_-1_to_1

D3mh4:
Issue : https://github.com/ossia/GBAP/issues/26


D3mh4:
The version that has been pushed is not yet operational. It correctly generates positions in a grid from -1 to 1, but the cursor detection area is not well defined. It would be necessary to allow the user to choose between a grid from -1 to 1 or a grid from 0 to 1 depending on their needs.

PR #1827, opened 2025-07-13
feature/enable vulkan raytracing extensions

YunmaoLeo:
ScreenNode: add vulkan device extensions when initializing vulkan instances

### DIFF
--- a/src/lib/score/gfx/Vulkan.cpp
+++ b/src/lib/score/gfx/Vulkan.cpp
@@ -1,9 +1,8 @@
 #include <score/gfx/Vulkan.hpp>
 
 #if defined(QT_FEATURE_vulkan) && QT_CONFIG(vulkan) && __has_include(<vulkan/vulkan.h>)
-#include "QRhiGles2.hpp"
-#include <QVulkanInstance>
-
+#include <rhi/qrhi.h>
+#include <QtGui/qvulkanfunctions.h>
 #include <mutex>
 namespace score::gfx
 {
@@ -236,8 +235,5 @@ QRhiVulkanNativeHandles* createRayTracingRhi(QVulkanInstance* inst)
 
   return handles;
 }
-
-
-
 }
 #endif
--- a/src/lib/score/gfx/Vulkan.cpp
+++ b/src/lib/score/gfx/Vulkan.cpp
@@ -1,6 +1,7 @@
 #include <score/gfx/Vulkan.hpp>
 
 #if defined(QT_FEATURE_vulkan) && QT_CONFIG(vulkan) && __has_include(<vulkan/vulkan.h>)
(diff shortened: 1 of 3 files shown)


jcelerier:
Change looks good but what happens if the user does not have a device that supports raytracing ? will it just safely ignore the changes?

YunmaoLeo:
@jcelerier 

> Change looks good but what happens if the user does not have a device that supports raytracing ? will it just safely ignore the changes?

Yes it will be safe. According to [Qt documententation](https://doc.qt.io/qt-6/qrhivulkaninitparams.html?), it is safe to add deviceExtensions even if the physical device might not support them.

> QRhiVulkanInitParams::deviceExtensions
> Optional, empty by default. The list of Vulkan device extensions to enable. Unsupported extensions are ignored gracefully.

Also, I've moved the code for creating the vulkan device and enabling raytracing extensions to `vulkan.cpp`. Please review the commit when you get a chance. Thank you!

jcelerier:
thanks! will take some time to review it today

src/lib/score/gfx/Vulkan.cpp
```diff
@@ -55,5 +56,188 @@ QVulkanInstance* staticVulkanInstance(bool create)
 
   return g_staticVulkanInstance;
 }
+
+// Returns true if the device supports ray tracing
```
jcelerier, line 59: All the vulkan code should be in the 

    #if defined(QT_FEATURE_vulkan) && QT_CONFIG(vulkan) && __has_include(<vulkan/vulkan.h>)

guard as some platforms don't have vulkan

src/lib/score/gfx/Vulkan.cpp
```diff
@@ -1,6 +1,7 @@
 #include <score/gfx/Vulkan.hpp>
 
 #if defined(QT_FEATURE_vulkan) && QT_CONFIG(vulkan) && __has_include(<vulkan/vulkan.h>)
+#include "QRhiGles2.hpp"
```
jcelerier: I don't think this should go in here, and it should be with <qrhigles2.hpp> header formata

jcelerier (changes requested):

src/lib/score/gfx/Vulkan.cpp
YunmaoLeo: Sorry, I've made a mistake here. I've changed it to correct header.

src/lib/score/gfx/Vulkan.cpp
YunmaoLeo: I am not really sure, but I think all the vulkan related code here is fully wrapped in the guard.
Let me know if I missed anything, happy to adjust further.

PR #1866, opened 2025-09-11
feature/buffer geometry

jcelerier:
- **process library: small performance optimization**
- **isf: simplify textures by using textureSize, now supported everywhere it matters**
- **cmake: also allow disabling plugins with SCORE_DISABLE_PLUGINS**
- **isf: deep rework of audio textures**
- **audio textures: small fixes**
- **isf: fix inlet names**
- **isf: introduce a new shader type: compute shader format**
- **core: minor code quality improvements**
- **vsa: fix serialization**
- **dataflow: prevent constant redraw hogging performance due to execution % bar in nodal view**
- **scenario: add new ux features**
- **gpu: allow geometry to hold gpu handles in addition to cpu buffers; add a possibility to specify texture format dynamically**
- **3d: add an array to texture object**
- **gpu: rework renderlist core loop to start enabling non/texture-io objects**
- **ci: missing header**
- **isf, csf, vsa: refactors and bugfixes**
- **renderer: ensure we do not do a resource update batch with nothing**
- **cables: prevent cables going from a slot outside of the nodal of an interval to outside the clipping zone**
- **model displau: compat with older builds**
- **event: take parent speed into account to avoid offset miscalculation in exec_state_facade::timings**
- **dataflow: improvements for rendering of objects ; only show a small dotted line when both ports aren't visible**
- **gpu: fix a resource being released twice in compute renderer**
- **ui: fix drops getting stuck due to a Qt bug**
- **gpu: introduce a buffer-to-geometry object**

### DIFF
--- a/src/plugins/score-plugin-gfx/CMakeLists.txt
+++ b/src/plugins/score-plugin-gfx/CMakeLists.txt
@@ -109,6 +109,11 @@ set(HDRS
     Gfx/Text/Process.hpp
     Gfx/Text/Layer.hpp
 
+    Gfx/BufferGeometry/Metadata.hpp
+    Gfx/BufferGeometry/Process.hpp
+    Gfx/BufferGeometry/Executor.hpp
+
+    Gfx/Graph/BufferGeometryNode.hpp
     Gfx/Graph/CommonUBOs.hpp
     Gfx/Graph/CustomMesh.hpp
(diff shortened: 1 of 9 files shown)


jcelerier:
redone in https://github.com/ossia/score/pull/1906

PR #1883, opened 2025-10-16
[commandStack] fix #1879

### LINKED ISSUE #1879: Crash on caling Score.undo() in the console
It seems to only be missing a ```caUndo``` check.

Adding this in src/plugins/score-plugin-js/JS/Qml/EditContext.cpp fixes it 

```c++
void EditJsContext::undo()
{
  auto doc = ctx();
  if(!doc)
    return;

  auto& s{doc->document.commandStack()};
  if (s.canUndo()) s.undo(); 
}
```

It would be a quick fix, butt maybe I am missing something.

thibaudk:
As mentioned in #1879


jcelerier:
all good thanks!

PR #1884, opened 2025-10-21
modify appimage.yml to avoid building older versions for CI/CD

edumeneses:


jcelerier:
I applied the patch to all the CI options, thanks for the patch!

jcelerier:
hmmm for some reason since this patch the "continuous" release log seems to be in append mode :D 

https://github.com/ossia/score/releases/tag/continuous

jcelerier:
@edumeneses do you think you'd have time to look into why this happened ?

edumeneses:
Still need this @jcelerier ?

PR #1902, opened 2025-11-18
feature/geometry filter in gpu thread

jcelerier:
- **ui: enforce double precision for spinbox range computation, fixes a graphical bug with undo and reloading**
- **geometry filter: put it in the gpu thread**


jcelerier:
merged in master manually

Issue #1905, opened 2025-11-25
Crash when manipulating Images list


edumeneses:
This crash happens when I send OSC messages to the Images process with new lists of image paths. It is an intermittent bug, but it happens consistently after some iterations.

```bash
Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0  ossia-score 0x000059017693f02c llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 44
1  ossia-score 0x000059017693c54c llvm::sys::RunSignalHandlers() + 188
2  ossia-score 0x000059017693fda9
3  libc.so.6   0x00007f0c1ac45330
4  libc.so.6   0x00007f0c1acadd55 __libc_free + 37
5  ossia-score 0x000059017b84d445 QRhiPassResourceTracker::~QRhiPassResourceTracker() + 149
6  ossia-score 0x000059017b84cfb8 QGles2CommandBuffer::resetCommands() + 328
7  ossia-score 0x000059017b84d023 QGles2CommandBuffer::resetState() + 51
8  ossia-score 0x000059017b83b178 QRhiGles2::beginFrame(QRhiSwapChain*, QFlags<QRhi::BeginFrameFlag>) + 104
9  ossia-score 0x000059017b6ce3cb QRhi::beginFrame(QRhiSwapChain*, QFlags<QRhi::BeginFrameFlag>) + 267
10 ossia-score 0x000059016ffeec1a score::gfx::Window::render() + 154
11 ossia-score 0x000059016ffef28c score::gfx::Window::event(QEvent*) + 620
12 ossia-score 0x000059017a365efe QApplicationPrivate::notify_helper(QObject*, QEvent*) + 318
13 ossia-score 0x000059017a366f0b QApplication::notify(QObject*, QEvent*) + 459
14 ossia-score 0x000059016f87bc63 SafeQApplication::notify(QObject*, QEvent*) + 211
15 ossia-score 0x000059017c6b66b8 QCoreApplication::sendEvent(QObject*, QEvent*) + 152
16 ossia-score 0x000059017b547e28 QPlatformWindow::deliverUpdateRequest() + 120
17 ossia-score 0x000059017b546f27 QPlatformWindow::windowEvent(QEvent*) + 39
18 ossia-score 0x000059017ba2dbc5 QXcbWindow::windowEvent(QEvent*) + 229
[...]

PR #1908, opened 2025-12-08
Draft: Fix memory leak in ImagesNode by replacing raw pointers with shared_ptr

edumeneses:
Addresses issue #1905 where manipulating the image list caused memory leaks due to improper manual cleanup of raw pointers.

### DIFF
--- a/tools/developer.sh
+++ b/tools/developer.sh
@@ -197,7 +197,6 @@ elif [[ "$OSTYPE" == "linux-gnu"* ]]; then
 (
   echo "[developer.sh] Installing dependencies"
   detect_linux_distro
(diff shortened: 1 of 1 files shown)


jcelerier:
the commits seems to have disappeared from the PR ? 🙂

edumeneses:
Yes, that was a dead end and didn't fix the problem. The real issue seems to come either from node parenting done wrong (assigning parenting to itself for some weird reason) and/or something between my object loader and model display.

I need to further investigate. Anyway, the Images process looks clean

edumeneses:
I can create another PR once I advance

PR #1912, opened 2025-12-15
fix/audio device reload

jcelerier:
audio: fix audio device document reload not recreating custom nodes


jcelerier:
cherry-picked to master in https://github.com/ossia/score/commit/5c495416b399776b8a662e41d4f72d1fed1ceb67

PR #1928, opened 2026-01-09
update to sh4lt 0.2.0

nicobou:


nicobou:
salut ! so it builds, but after some testing, there is at least on fix to add. coming soon :p

jcelerier:
Ah merci pour le patch! Looks like it's just macos which does not support a flag: 

>Users/runner/work/score/score/3rdparty/sh4lt/sh4lt/ipcs/unix-socket-client.cpp:56:88: error: use of undeclared identifier 'MSG_CONFIRM'
>   56 |             socket_.fd_, &proto_->quit_msg_, sizeof(proto_->quit_msg_), MSG_NOSIGNAL | MSG_CONFIRM);

Maybe it can be enabled only if(LINUX/BSD)?

nicobou:
Yes, Sh4lt now builds only on Linux (I think, but not sure since I could not test on OSX). 
Also, I tested the update with GStreamer pipelines. So unless you see something else, it is ready for a merge.
Along this PR, an other PR in the Ossia doc fixes the GStreamer examples : https://github.com/ossia/score-docs/pull/83

And thank @jcelerier for the great work !

PR #1933, opened 2026-01-11
feature/codeeditor dock

jcelerier:
- **window: make the right side also a tab list**
- **script: dock script editors inside the right pane**
- **gfx: use default prettyName implementation**


jcelerier:
@samamou with this PR code editing looks like this:

https://github.com/user-attachments/assets/bc38d6bc-b8b8-4255-8f1a-bbb93b8fd27d


I believe it's an improvement but I'd like your input

jcelerier:
@lou-evoy do you have some input on this ?

samamou:
I prefer this over opening in the separate window 

but my usage is really opening in an external editor so i can't give you much :/

PR #1975, opened 2026-02-28
audio: add a native ASIO backend on Windows

### LINKED ISSUE #1422: Use ASIO directly instead of PortAudio on Windows
It causes crashes at startup if there is a buggy driver due to mandatory initialization in Pa_Initialize();

jcelerier:
Adds a native ASIO audio backend on Windows, using the Steinberg SDK directly
rather than going through PortAudio's ASIO host.

**Depends on ossia/libossia#920.** The engine itself (`ossia::asio_engine`) lives
in libossia; that PR fixes the driver enumeration and driver-ownership bugs this
backend cannot work without, and adds the `active_driver()` /
`control_panel_result` API used here. The submodule currently points at
`ossia/libossia@11300f2a` on the `feature/asiosdk` branch — **it needs
re-pointing at the merged commit on libossia `master` before this is merged.**

## What it adds

- `Audio/ASIOInterface.hpp` — `NativeASIOFactory`, listed as "ASIOSDK" alongside
  the existing PortAudio-based "ASIO" backend.
- Device list with channel counts, a control-panel button and a rescan button.
- `SCORE_AUDIO_BACKEND=asio` to force the backend at startup.
- `Audio/AudioApplicationPlugin.{hpp,cpp}` — `with_engine_stopped()`, a
  backend-agnostic helper described below.

Guarded by `OSSIA_ENABLE_ASIO` (on by default on Windows in libossia) and
`TARGET asio::sdk`, so nothing changes on other platforms or without the SDK.
Diagnostics go through `ossia::logger()`, so they appear in the Messages panel
as well as on a console.

## Things found during bring-up

**Unusable devices are labelled, not hidden.** A driver whose hardware is
unplugged still enumerates. Previously selecting it just failed with the generic
"the desired audio settings could not be applied", so the reason is now shown in
the list: `Audio 8 DJ (device not connected)`, mapped from `ASIOInit`'s error
code. `ASIOCard::name` stays the verbatim driver name — it is the identity stored
in the settings and handed to `asio_engine` — and only the combo box label is
decorated.

**Rescanning cannot happen behind the engine's back.** Probing a driver for its
channel counts means loading it, and `AsioDrivers::loadDriver()` releases
whatever driver is currently loaded, which would cut a streaming engine off from
its hardware. Since `rescan()` runs on every `setDriver()`/`initDriver()`,
reopening the audio settings during playback used to do exactly that. Probing is
now skipped while an ASIO engine streams.

**Startup cost.** `rescan()` is reached three times while score starts up — this
factory's constructor, `Model::initDriver` and `ApplicationPlugin::initialize`,
all through the shared `AudioFactory::initialize()` API. Listing drivers is a
cheap registry walk (~11 ms), but each probe loads the driver's DLL through COM:
measured 22–32 ms per driver, before real hardware even answers. Doing all of it
three times cost roughly 330 ms before the GUI could paint. Split into a cheap
idempotent listing plus a probe that runs at most once per driver:

| | Enumerations | Driver DLL loads | Measured |
|---|---|---|---|
| Before | 3 | 9 | ~331 ms |
| After | 1 | 3 | ~111 ms |

**Refreshing the list is on the show event, not `make_settings()`.** score builds
every settings widget during startup, so refreshing at construction just puts the
probe back on the startup path.

**Rescan button.** When nothing holds an ASIO driver — no engine, or another
backend — it probes straight away and leaves playback alone. When an ASIO device
is in use it goes through `ApplicationPlugin::with_engine_stopped()`, which stops
the engine, probes, and brings it back; the tooltip warns that this interrupts
playback. That helper is backend-agnostic and reusable by any driver needing
exclusive hardware access, which is why it lives in the application plugin rather
than here.

## Two deliberate behaviours, in case they look like bugs

- `available()` returns true even with no ASIO drivers installed — the backend is
  meant to stay visible in the settings regardless.
- `initialize()` auto-selects the first device without checking connectivity — on
  the common single-driver system this picks the right one on first run.

## Testing

On Windows (clang64) with three installed ASIO drivers: ASIO4ALL v2, NI Audio 8
DJ, NI Traktor Kontrol S4 MK2.

- Playback verified with real hardware through the Traktor Kontrol S4 MK2.
- Repeated Settings > Audio > OK cycles keep the engine alive.
- Control panel opens the selected driver, and explains itself when another
  driver is streaming instead of showing the wrong panel.
- Disconnected devices labelled; hot-plug picked up on reopening the settings.
- Rescan covered in all four states: engine stopped, ASIO streaming, ASIO
  selected while another backend still runs, and device unplugged mid-playback.

## Note for reviewers

`2d4a2ee` is the original backend commit rebased onto current `master`; the seven
commits after it are the bring-up fixes. It no longer bumps the submodule, since
`master` already pointed at the libossia commit it needed.

## Disclosure

### DIFF
--- a/src/plugins/score-plugin-audio/Audio/ASIOInterface.hpp
+++ b/src/plugins/score-plugin-audio/Audio/ASIOInterface.hpp
@@ -1,0 +1,204 @@
+#pragma once
+#include <ossia/detail/config.hpp>
+#include <Audio/AudioInterface.hpp>
+#include <Audio/Settings/Model.hpp>
+#include <Audio/Settings/View.hpp>
+
+#include <score/command/Dispatchers/SettingsCommandDispatcher.hpp>
+#include <score/tools/Bind.hpp>
+#include <score/widgets/SignalUtils.hpp>
+
+#include <QComboBox>
+#include <QFormLayout>
+#include <QPushButton>
+#include <QLabel>
+
+#if defined(OSSIA_ENABLE_ASIO)
+#include <ossia/audio/asio_protocol.hpp>
+
+extern AsioDrivers* asioDrivers;
+namespace Audio
+{
+struct ASIOCard
+{
+  QString name;
+  int driver_index{-1};
+  int inputChan{};
+  int outputChan{};
+};
+
+class NativeASIOFactory final
+    : public QObject
+    , public AudioFactory
+{
+  SCORE_CONCRETE("2d21a3aa-f108-4e05-a3a0-8fd6b150bda5")
+public:
+  std::vector<ASIOCard> devices;
+
+  NativeASIOFactory() { rescan(); }
+  ~NativeASIOFactory() override { }
+
+  bool available() const noexcept override { return !devices.empty(); }
+
+  void
+  initialize(Audio::Settings::Model& set, const score::ApplicationContext& ctx) override
+  {
+    rescan();
+
+    if(!set.getCardOut().isEmpty())
+    {
+      // Check that the saved device still exists
+      auto it = std::find_if(
+          devices.begin(), devices.end(),
+          [&](const ASIOCard& d) { return d.name == set.getCardOut(); });
+      if(it != devices.end())
+        return;
+    }
+
+    // Auto-select first device
+    if(devices.size() > 1)
+    {
+      auto& dev = devices[1]; // [0] is "No device"
+      set.setCardIn(dev.name);
+      set.setCardOut(dev.name);
+      set.setDefaultIn(dev.inputChan);
+      set.setDefaultOut(dev.outputChan);
+    }
+  }
+
+  void rescan()
+  {
+    devices.clear();
+    devices.push_back(ASIOCard{QObject::tr("No device"), -1, 0, 0});
+
+    try
+    {
+      auto cards = ossia::asio_engine::enumerate_drivers();
+      for(auto& card : cards)
+      {
+        // To get channel counts, we must briefly load each driver
+        int ins = 0, outs = 0;
+        if(loadAsioDriver(const_cast<char*>(card.name.c_str())))
+        {
+          ASIODriverInfo info{};
+          info.asioVersion = 2;
+          if(ASIOInit(&info) == ASE_OK)
+          {
+            long numIn = 0, numOut = 0;
+            ASIOGetChannels(&numIn, &numOut);
+            ins = (int)numIn;
+            outs = (int)numOut;
+            ASIOExit();
+          }
+          if(asioDrivers)
+            asioDrivers->removeCurrentDriver();
+        }
+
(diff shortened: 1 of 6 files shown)


jcelerier:
@ogauthiersat are you down for trying to rebase, test and finish that one?

ogauthiersat:
> @ogauthiersat are you down for trying to rebase, test and finish that one?

I rebased the work you did on latest master.

I had to bring another PR to libossia as ASIO did not list any drivers because of an UNICODE/ANSI issue and fix a an driver release issue on audio engine restart. https://github.com/ossia/libossia/pull/920

I tested this using ASIO4ALL and an 6 in / 4 out NI audio interface.

Devices are scanned at startup, re-scanned on audio setting window if the ASIO driver is not yet running. I added a "Rescan" button to the device list that trigger an audio engine restart if an driver is currently streaming (ASIOSDK cannot probe driver status (device connected and channel count) if a driver is already loaded).

jcelerier:
works great on my two ASIO devices (FAS Audio Axe-Fx III through its own asio driver, and B&W MM-1 through asio4all) here! I'll try with more later on, at the SAT.
Could you add the ASIO GPL license file here: https://github.com/ossia/score/blob/192676568c8073a83a2d1e54aacef87afdaf5a65/src/lib/CMakeLists.txt#L701 so that it show up with the other licenses in the About dialog ?

ogauthiersat:
> works great on my two ASIO devices (FAS Audio Axe-Fx III through its own asio driver, and B&W MM-1 through asio4all) here! I'll try with more later on, at the SAT. Could you add the ASIO GPL license file here:
> 
> https://github.com/ossia/score/blob/192676568c8073a83a2d1e54aacef87afdaf5a65/src/lib/CMakeLists.txt#L701
> so that it show up with the other licenses in the About dialog ?

I tried adding the sub-submodule license file, but it does not appears in the about dialog like so :

```
AddLicenseFile(license_text asio "${CMAKE_SOURCE_DIR}/3rdparty/libossia/3rdparty/asio/LICENSE.txt")
```

There's already an entry for ASIO old license that appears to be hard-coded in src\lib\core\presenter\AboutDialog.cpp:215.

ogauthiersat:
@jcelerier I rebased on latest master and managed to hook the license file into the about dialog correctly.

jcelerier:
hmm, trying with ASIO4ALL and my usb loudspeakers (MM-1) it seems to hang

ogauthiersat:
> hmm, trying with ASIO4ALL and my usb loudspeakers (MM-1) it seems to hang

I tested my msys/clang build along with the sdk build from the CI and cannot seem to reproduce any hang with ASIO4ALL on my side.

jcelerier:
thanks, I'll investigate more on my windows box!

PR #1995, opened 2026-03-15
UI: Increase hitboxes and interaction areas for core elements

### LINKED ISSUE #1005: (2.5.2) Displacement/stiching: process upper left corner basic actions hard to perform
I find a bit painful the user experience of simple actions such as displacing a process or attaching somewhere, setting its start and stop position relatively to other elements etc: basically everything that involves drag&drop is for me confusing and time-consumming. I would expect things to "snap" on to an other much more easily:

- case 1: I have two independent processes that are linked to nothing else. I would like that dropping one near the end of the other one proposes to attach them automatically
- case 2: when I drag and drop a circle element that is fixed on a vertical axis, I would expect to create a connection, without clicking the + button

Then, for a single process, the upper left corner has too many control points, they are tiny and do not help understanding the action that they allow to perform. The pointer is also inconsistent (it sometimes changes even if we can't perform any action on the process, such as a hand appearing when hovering over elements that we can't drag) and do not follow what I would call "conventions": I think that the cross pointer is really awful, I would suggest only using it for drawing or selecting elements.

### LINKED ISSUE #1994: [Feature request] Enlarge hitboxes

edumeneses:
Closes #1994 

Increased hitboxes and interaction areas for various UI components to improve selection and drag-and-drop usability:
- Cables: Increased stroke width from 7 to 14.
- Intervals: Added a 5-pixel margin and increased drop threshold to 12.
- Conditions: Increased stroke width from 1 to 10.
- Events: Widened bounding box from 6 to 12.
- States: Enlarged visual radii (full: 6->8, point: 3.5->5).
- Ports: Expanded interaction area to 20x20.
- Slot Headers: Widened drag area and increased drop threshold between slots.

### DIFF
--- a/src/plugins/score-lib-process/Process/Dataflow/CableItem.cpp
+++ b/src/plugins/score-lib-process/Process/Dataflow/CableItem.cpp
@@ -113,7 +113,7 @@ static const QPainterPathStroker& cableStroker()
     QPen pen;
     pen.setCapStyle(Qt::PenCapStyle::RoundCap);
     pen.setJoinStyle(Qt::PenJoinStyle::RoundJoin);
-    pen.setWidthF(7.);
+    pen.setWidthF(14.);
     return pen;
   }()};
   return cable_stroker;
--- a/src/plugins/score-lib-process/Process/Dataflow/PortItem.cpp
+++ b/src/plugins/score-lib-process/Process/Dataflow/PortItem.cpp
@@ -568,8 +568,8 @@ void PortItem::setHighlight(bool b)
 
 QRectF PortItem::boundingRect() const
 {
-  constexpr auto max_diam = 13.;
-  return {0., 0., max_diam, max_diam};
+  constexpr auto max_diam = 20.;
+  return {-3.5, -3.5, max_diam, max_diam};
 }
 void PortItem::paint(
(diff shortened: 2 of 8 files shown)


jcelerier:
works here! but for me it seems to make it harder to select the cross to create new objects in the timeline: basically only the lower-right or upper-left corner triggers the "creation" yellow or red plus signs now :/

https://github.com/user-attachments/assets/559a84c5-9fb6-4b6c-bbc1-4f604a873b1b

jcelerier:
the correct thing to do here in addition to this change is: in src/plugins/score-plugin-scenario/Scenario/Document/State/StateView.hpp, we should implement the following virtual methods for QGraphicsItem, so that it knows that the item does not only have a square shape:

    virtual QPainterPath opaqueArea() const;
    virtual bool contains(const QPointF &point) const;
    virtual QPainterPath shape() const;

an example is in src/plugins/score-plugin-curve/Curve/Segment/CurveSegmentView.hpp.

Generally most items should implement these virtual methods, it's more work but also means that if you click outside of the visible area it will be correct

edumeneses:
ops, this fell into the cracks..... Getting back to it asap

edumeneses:
Should work fine now. Can you please check @jcelerier ?

jcelerier:
thanks !!

PR #2001, opened 2026-03-18
linux: fix missing avahi-daemon program warning on some distro

ogauthiersat:


jcelerier (approved):
thanks for nailing this one down!

PR #2002, opened 2026-03-18
tools: fix missing lf in mac custom app launcher environment

ogauthiersat:


jcelerier (commented):
hmm  what does this change ? the patch seems to only add whitespace

ogauthiersat:
> hmm what does this change ? the patch seems to only add whitespace

Yes, it adds line feeds, similar to how the launcher script is done in the linux package.

Without extra LFs, if you input an environment file that does not end with a trailing LF, you end up with a non working launcher like such :

```
export DOMEPORTPRO_BASIC=1exec "$SCRIPT_DIR/app-bin"          --ui "${RESOURCES_DIR}/qml/Main.qml"     "${RESOURCES_DIR}/app.score"     "$@"
```

jcelerier:
ahhh thanks for the explanation! I would have thought that there would be one after the LAUNCHER_EOF, good to know

jcelerier (approved):

PR #2010, opened 2026-03-29
media handlers: greatly improve ffmpeg integration, integrate gstreamer and gphoto2

jcelerier:
- **protocols: separate video in, video out and media categories**
- **device dialog: sort things**
- **gfx: support more cases of parameters in gfx devices**
- **gfx: add yuv encoding shaders**
- **audio: expose execution_samples as a global for inter-thread sync**
- **libav: rework libav device completely. Add input support and presets.**
- **gstreamer: add support for using gstreamer I/O as a device**
- **gphoto: add support for using gphoto cameras if installed**


jcelerier:
example: `gstreamer testsrc` -> `ffmpeg xv output` and `ffmpeg testsrc` -> `gstreamer ximagesink`. We doin' crimes :D 

https://github.com/user-attachments/assets/27d3e0e6-e238-4b18-b5d6-456a561c29f0

jcelerier:
In terms of design: gstreamer symbols are loaded dynamically. My idea is that score should continue shipping with a minimal ffmpeg. For extended codec support, we could ship a pre-compiled .so / .dylib / .dll of gst + codecs + ..., as an extension in the package manager. This way, Windows and macOS users who don't need extensive video I/O capabilities won't have to download an additional 150 megabytes, and linux desktop users who have GStreamer already installed will directly benefit since it will use their distro version. Flatpak users and users who build from source with their distro also benefit of whatever codecs their distro provide for ffmpeg (which unlike gstreamer is monolithic).

TODO: one thing that we absolutely have to do is make sure that we can blacklist plug-ins from being loaded as GStreamer has a couple plug-in that integrates with Qt ( libgstqml6.so libgstqmlgl.so here in /usr/lib/gstreamer-1.0). If they get loaded they will in turn load the distro Qt which will definitely crash in the AppImage situation since it may not be the same Qt version.

jcelerier:
@edumeneses @ogauthiersat this will also allow recording & saving media, which we can use for instance for our dome format conversion use cases

PR #2011, opened 2026-03-29
feature/window capture

jcelerier:
- **protocols: separate video in, video out and media categories**
- **gfx: add a window capture device**
- **gfx: window capture: extend it to work with full-screen and screen portions if available**

### DIFF
--- a/ci/debian.bookworm.deps.sh
+++ b/ci/debian.bookworm.deps.sh
@@ -27,6 +27,7 @@ $SUDO apt-get install -qq --force-yes \
      libsuil-dev liblilv-dev lv2-dev \
      libdbus-1-dev \
      libdrm-dev libgbm-dev \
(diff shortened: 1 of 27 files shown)


jcelerier:
@edumeneses this one allows to grab a specific window, screen region, etc. in the most efficient way possible :-)

edumeneses:
On my machine (Ubuntu 24.04 - Wayland), I can grab other windows and score's own window (although it doesn't list all of them, even after refreshing).

Region, Single-screen, and All-screen modes produce full black.

jcelerier:
thanks for the test!

jcelerier:
on X11 they work but on wayland I think it depends on the compositor..

jcelerier:
Tried to push a fix for this on wayland

PR #2025, opened 2026-04-08
linuxcheck: display the packages to install

jcelerier:


ogauthiersat:
I tested the AppImage build 8f8070b on multiple live CDs, Ubuntu 24.04, 26.04, Debian 13 KDE, Manjaro KDE and Rocky 10 Gnome.

The only distro that needed extra packages was Ubuntu 26.04. The suggested command `sudo apt install libopengl0 libxcb-xkb1 libxkbcommon-x11-0` worked as expected.

All distros were using Wayland by default.

Most distros were using zenity, the only exception was Debian KDE where its using kdialog.

All distros gave warnings for missing groups, albeit, I tested audio with `faust-synthesis.score` example successfully even if `audio` group was missing.

Here's the list of default groups :

```
Ubuntu: ubuntu adm cdrom sudo dip plugdev users lpadmin sambashare
Debian: user cdrom floppy sudo audio dip video plugdev users scanner bluetooth
Manjaro: manjaro sys network power lp wheel
Rocky: liveuser wheel
```

I suggest removing group checks and warnings for now because most users will have to click through a warning at startup, which is annoying and confusing. It would be better to document special device access configurations in the manual pages for such devices, or maybe add runtime checks.

I also encountered an issue with zenity on Ubuntu 26.04 while running on a small screen resolution (1280x800). The warning text message pushes the OK button out of screen because zenity default width is too narrow.

Screenshot_20260409_103229_26 04

I tested with `width=800` where it renders correctly. Here's the command modification suggestion :

https://github.com/ossia/score/blob/8f8070ba24006b007934ca8940636ff066006caf/src/linuxcheck/linuxcheck.cpp#L56

```
    if(!system(fmt::format("zenity --width=800 --warning --text=\"{}\"", text).c_str()))
```
Screenshot_20260409_104345_zenity_800px

jcelerier:
removed the group checks, will need to embed them into the device dialogs

jcelerier:
thanks for the width=800 suggestion!

ogauthiersat:
Tested build d1d6b25. I get an empty warning popup. Warning text appears that is never empty as a space character is always appended on[ line 26](https://github.com/ossia/score/blob/feature/linuxcheck-package/src/linuxcheck/linuxcheck.cpp#L26).

```
Welcome to ossia score!

All the required software or system features are not installed, and thus the software cannot run properly.
Make sure that the following software, features, configuration and libraries are available on your system: 
 
 

If you do not want to see this message again, run the app with SCORE_SKIP_LINUXCHECK=1 environment variable set.
```

Screenshot_20260410_115048

jcelerier:
wops, that's me pushing my debug checks... should be good now, merging!

PR #2030, opened 2026-04-16
ui: fix HiDPI rendering issues introduced by the vector-gui commit

bltzr:
Two bugs in the vector_gui rendering path on Retina displays:

1. SimpleTextItem and TemporalIntervalHeader were drawing text using m_line.height() / m_pixmap.height() (physical pixels) as the baseline Y coordinate in logical scene units, causing labels to render below their bounding rect and misaligning hit areas. Fixed by using m_rect.height() / m_textRectCache.height() instead.

2. get_cursor() was setting devicePixelRatio before loading the pixmap (which resets it), and was doubling the hotspot coordinates for @2x cursors. QCursor hotspot is in logical pixels, so doubling it placed the active point at the bottom-right of the cursor image. Fixed by loading before setting dpr, and not doubling the hotspot.

### DIFF
--- a/src/lib/score/graphics/TextItem.cpp
+++ b/src/lib/score/graphics/TextItem.cpp
@@ -55,7 +55,7 @@ void SimpleTextItem::paint(
       painter->setPen(m_color->pen1);
     painter->setFont(m_font);
     painter->setBrush(skin.NoBrush);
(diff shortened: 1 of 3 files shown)


jcelerier (approved):
Thanks!

PR #2033, opened 2026-04-20
sound: add video file support to sound process

ogauthiersat:
Allow sound process to load video file first audio track at creation using JS (ie. `Score.createProcess(Score.rootInterval(), "Sound file", "video.mp4")`).

Allow loading video file first audio track by drag'n'dropping a video file into an existing sound process.

Fix sound drop regex to match extension as suffix only.


jcelerier:
looks good thanks!!! I'll merge it, I think that in a later time it'd be good to refactor the duplicated supported formats and also take into account the supported formats dynamically - right now it lists the formats supported by the official builds, but on distro versions it may vary at runtime depending on what the distro built their ffmpeg with

PR #2052, opened 2026-05-21
1886 feature request mouse selection and dragging of multiple processes

### LINKED ISSUE #1886: Feature request: mouse selection and dragging of multiple processes
This is a quality-of-life feature request:

- Choosing multiple objects with CRLT + left click and drag (it already works with CTRL + click on each object, but it isn't very pleasant)
- Drag multiple objects: this doesn't work at all (at least on Linux). Organizing scores is painful without this feature

Those features will also be helpful when the user wants to convert between Internals, Nodes, and Scenarios.

edumeneses:
Done with Claude

Closes #1886

### DIFF
--- a/src/plugins/score-lib-process/Process/Dataflow/NodeItem.cpp
+++ b/src/plugins/score-lib-process/Process/Dataflow/NodeItem.cpp
@@ -942,10 +942,6 @@ enum Interaction
 } nodeItemInteraction{};
 QSizeF origNodeSize{};
 bool nodeDidMove{};
-// When pressing on an already-selected node, the selection change is deferred
-// to the release event so that the whole selection can be dragged. The
-// toggle/collapse only happens on release if no drag occurred. The modifier
-// state at press time is captured for that deferred handling.
 bool nodeSelectOnRelease{};
 bool nodeSelectCumulation{};
 }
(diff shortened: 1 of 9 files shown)


edumeneses:
@jcelerier thanks for the rebase — I pushed two more commits on top of it (branch stays rebased on current master, fast-forward):

- `4cc6b3502` **nodal: defer node selection change to release** — pressing an already-selected node ran `filterSelections` immediately on press. With Ctrl held (required for box-select) it toggled the clicked node off, so dragging moved every selected node *except* the clicked one; without a modifier it collapsed the selection to the single clicked node, making multi-drag impossible. Now the selection change is deferred to release: a drag moves the whole selection untouched, while a click without drag keeps the standard behaviour (collapse to clicked, or toggle off with Ctrl).

- `a0683cb2e` **scenario: support deleting a multi-selection of processes** — `ScenarioEditor::remove` only handled the single-element case for processes, so Delete on a nodal box-selection removed nothing/just one. Added a branch that removes each selected process from its interval in a single macro, via a new `RemoveMultipleProcessesFromInterval` aggregate command.

Both build cleanly here and the box-select → drag and multi-delete flows now work. Let me know if you'd prefer the multi-delete handled elsewhere than `ScenarioEditor::remove`.

PR #2053, opened 2026-05-22
nodal: persist zoom level when navigating in and out of intervals

### LINKED ISSUE #1961: [Cosmetic feature] persistent zoom setting
This is 100% cosmetic UX feature that helps when presenting a score, but it is not high-priority.

When organizing scores, sometimes users (aka me) want to have a TD-like visualization (specifically useful for node-based devices (scores acting as state-machines).

One way to achieve this is to use scenarios as subpatches (or sub-scores 😉). It requires a bit of config to make scenario nodes run indefinitely, but it works. What I want to achieve is something like the first image below:

Image

Unfortunately, the zoom level resets once one enters and leaves the scenario:

Image

The expected behaviour would be to have a persistent zoom level on scenarios and nodes. That would also allow us to make specific process parameters from a sub-scenario always visible for interaction, e.g., having the image position visible to interact with:

Image

edumeneses:
Dome with Claude

IntervalModel already stored m_nodalScale (serialized) but the view never wrote to it. Now wheelEvent, zoomTo, recenter, and rescale all call setNodalScale, and the constructor restores the saved value before the deferred recenterRelativeToView fires.

### DIFF
--- a/src/plugins/score-plugin-scenario/Scenario/Document/Interval/FullView/NodalIntervalView.cpp
+++ b/src/plugins/score-plugin-scenario/Scenario/Document/Interval/FullView/NodalIntervalView.cpp
@@ -101,6 +101,12 @@ NodalIntervalView::NodalIntervalView(
       }
     }
   }
+  if(const double savedScale = m_model.nodalScale(); savedScale != 1.0)
+  {
+    m_container->setScale(savedScale);
+    m_zoomLevel = std::log(savedScale) / std::log(g_zoom_base);
+  }
+
   QTimer::singleShot(1, this, &NodalIntervalView::recenterRelativeToView);
 }
 
@@ -157,6 +163,8 @@ void NodalIntervalView::recenter()
   auto delta = ourCenter - childCenter;
 
   m_container->setPos(delta);
+  m_zoomLevel = std::log(z) / std::log(g_zoom_base);
+  const_cast<IntervalModel&>(m_model).setNodalScale(z);
   const_cast<IntervalModel&>(m_model).setNodalOffset(QPointF{});
 }
 
@@ -167,6 +175,7 @@ void NodalIntervalView::rescale()
   auto childRect = enclosingRect();
 
   m_container->setScale(1.0);
+  m_zoomLevel = 0.;
 
   auto childCenter
       = m_container->mapRectToParent(childRect).center() - m_container->pos();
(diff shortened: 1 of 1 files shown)


edumeneses:
Closes #1961

src/plugins/score-plugin-scenario/Scenario/Document/Interval/FullView/NodalIntervalView.cpp
```diff
@@ -101,6 +101,12 @@ NodalIntervalView::NodalIntervalView(
       }
     }
   }
+  if(const double savedScale = m_model.nodalScale(); savedScale != 1.0)
+  {
+    m_container->setScale(savedScale);
+    m_zoomLevel = std::log(savedScale) / std::log(g_zoom_base);
```
jcelerier, line 111: must check that g_zoom_base > 1. otherwise this can cause NaN / division by zero

src/plugins/score-plugin-scenario/Scenario/Document/Interval/FullView/NodalIntervalView.cpp
```diff
@@ -157,6 +163,8 @@ void NodalIntervalView::recenter()
   auto delta = ourCenter - childCenter;
 
   m_container->setPos(delta);
+  m_zoomLevel = std::log(z) / std::log(g_zoom_base);
```
jcelerier, line 170: same here

jcelerier (commented):
some things to validate as they are the cases where this code used to break:

- switching across multiple objects between nodal view and timeline view
- opening multiple documents and switching around between tabs

thanks!!!

src/plugins/score-plugin-scenario/Scenario/Document/Interval/FullView/NodalIntervalView.cpp
edumeneses: Since g_zoom_base is constexpr, I enforced the invariant at compile time with static_assert(g_zoom_base > 1.0) next to its definition — guarantees std::log(g_zoom_base) is strictly positive, so no division by zero / NaN here. Fixed in 2a8d245b4.

src/plugins/score-plugin-scenario/Scenario/Document/Interval/FullView/NodalIntervalView.cpp
edumeneses: Same static_assert(g_zoom_base > 1.0) covers this call site too. Fixed in 2a8d245b4.

edumeneses:
I'll build and test locally for the mentioned break cases. Keep you posted

edumeneses:
It looks fine on my end. It doesn't break with multiple scores open and switching between processes, scores, and modes (timeline, nodal)

jcelerier (approved):

PR #2093, opened 2026-06-22
gfx: set Text process help URL

edumeneses:
The Text process had an empty manual_url, so pressing F1 opened nothing. Point it to the existing reference page.


jcelerier:
thanks !

PR #2110, opened 2026-07-12
avnd: support comboboxes with runtime-populated item lists

bltzr:
# PR (ossia/score): runtime-populated comboboxes for avnd objects

**Branch:** `feature/dynamic-combobox` (off `master`)
**Depends on:** avendish `feature/dynamic-combobox` (companion PR).
**Diff:** 4 files, +119/-10.

## What

Lets an avendish object repopulate a combobox port's item list at run time —
e.g. a "File" combobox that shows the audio files currently in a folder,
updating as the folder changes. Consumes the avendish `halp::dynamic_combobox`
port.

## Changes

- **`Process::ComboBox`** (`WidgetInlets.hpp/.cpp`): `setAlternatives()` +
  `alternativesChanged` signal. `setAlternatives` refreshes the value domain
  and re-clamps the current value when the list shrinks. Alternatives were
  already serialized, so documents reopen with their last-known items.
- **Widgets** (`ControlWidgets.hpp`): the inspector and canvas comboboxes read
  their items live and rebuild on `alternativesChanged` (previously the list
  was fixed at widget construction).
- **Executor** (`Crousti/Executor.hpp`): `connect_dynamic_items` injects the
  port's `update_items` callback, marshalling runtime updates to the model
  inlet on the main thread (`QPointer`-guarded). Nodes with type-only inputs
  are skipped.

## Compatibility

Purely additive. Objects and ports that don't use the feature are unaffected;
the widget/executor paths are guarded by `requires`/`if constexpr` on the new
members.

## Testing

A throwaway `DynComboTest` object (in the score-avnd-flucoma addon) lists a
folder's audio files in a `dynamic_combobox` and follows folder changes live;
verified in the running app. Not part of this PR.


bltzr:
Depends on the avendish port: celtera/avendish#171 (`halp::dynamic_combobox`). That should land first — this PR consumes its `update_items` callback.

bltzr:
Closing for now — the avendish port this consumes (celtera/avendish#171) has no real consumer yet, so parking both until one exists rather than adding to the review queue. Kept on `bltzr:feature/dynamic-combobox`; will reopen together with the avendish PR when adopted.

bltzr:
Reference, not an active review request — parked alongside the avendish port it consumes (celtera/avendish#171). Working example: [`DynComboTest`](https://github.com/bltzr/score-avnd-flucoma/blob/main/Flucoma/DynComboTest.hpp).

We solved our own case (Granola's sound picker) with a custom widget rather than this binding, so nothing depends on it. Happy to revive both together if you think a runtime-populated combobox is worth having generally.

jcelerier:
eventually it's a feature we want though ! it's very useful in many cases

bltzr:
OK then I'll reopen it and move this forward, as it's going to be useful for the flucoma port

bltzr:
Runtime-tested end-to-end now (populate at document load, save/reload round-trip, selection preserved). Two additions since last push:

- ComboBox now serializes its folder-port metadata, so a loaded document re-scans the folder and keeps live refresh (previously factory-set only, lost on deserialization).
- QGraphicsCombo: clicking a combobox on the canvas now opens a QMenu popup (items = current alternatives). Before, the only interaction was the drag-to-spin, which made comboboxes effectively read-only on the process panel while working in the inspector — this fixes that for every combobox/enum port, independently of this feature.

celtera/avendish#171 should still land first.

bltzr:
Follow-ups, all runtime-tested:

- Folder-backed comboboxes now install a QFileSystemWatcher, so files written while the document is open (an upstream analysis, a recorder…) appear in the list immediately.
- Value refresh keeps a selection that names a not-yet-existing file (it becomes selected once written) instead of clamping to an arbitrary entry.
- The sibling port may also hold a *file* path — the file's containing folder is listed then, so a picker can sit next to an existing sound-file port without a dedicated folder port.

Exercised across ~45 processes in a FluCoMa addon: full analysis chain (onset slicing → corpus building → KD-tree fit/query) runs zero-config with the pickers.

jcelerier:
Does this also support non-file/folder-backed comboboxes ? for instance custom presets, text input by the user etc.

bltzr:
yes, I think so

bltzr:
To be precise about it (firmer than my earlier reply): yes for any runtime-populated list. `halp::dynamic_combobox` is only the `update_items` callback — the object (or the binding) can push item lists from any source: preset names, labels computed from an analysis, results of a query… `folder_combobox` is just one edit-time producer layered on top; nothing in the binding assumes files. [`DynComboTest`](https://github.com/bltzr/score-avnd-flucoma/blob/main/Flucoma/DynComboTest.hpp) shows the generic path (object code repopulating the list at runtime).

What it does *not* cover is freeform text typed by the user — that's an *editable* combobox rather than a populated list. That'd be a small follow-up (an `editable` flag on the port + a line-edit in the popup); happy to add it to this PR or as a separate one if you want it.

Also just pushed to celtera/avendish#171: concepts relicensed like the rest of `avnd/concepts`, and `dynamic_items_parameter` is now shape-based (any function-like `update_items` member; no concrete `std::` types in the constraint).

jcelerier:
thanks, merged !

PR #2114, opened 2026-07-13
avnd ui: resolve file-path templates before syncing them to custom items

bltzr:
### Problem

File-chooser port values are stored in the document as path templates (`<PROJECT>:...`, `<LIBRARY>:...`, or document-relative). Score's own file widgets resolve them through `score::locateFilePath`, but the avnd layout builder pushed the **raw template string** into custom UI items' `value`.

So an add-on widget bound to a soundfile/file/folder port received a path it cannot use — e.g. a waveform widget that decodes the file to display it in edit mode could never find `<PROJECT>:samples/foo.wav`. Values worked during execution only because the engine resolves paths on its own side, which made panels that depend on the path (waveform displays, folder pickers) appear empty until playback started.

### Fix

Resolve through `score::locateFilePath` in `LayoutBuilder::setupControl`, both for the initial sync and in the `valueChanged` connections, whenever the inlet is a `Process::FileChooserBase` or `Process::FolderChooser` and the value is a non-empty string. Other value types and non-file ports are untouched.

Verified with an avnd add-on (Granola) whose custom waveform widget decodes the Sound port's file: with this change the widget receives `/abs/path/samples/foo.wav` at document load and can display the waveform in edit mode, matching the behavior of score's built-in file widgets.

https://claude.ai/code/session_01AvD2PFHZtW6pfsh3gGhayg

### DIFF
--- a/src/plugins/score-plugin-avnd/Crousti/Layer.hpp
+++ b/src/plugins/score-plugin-avnd/Crousti/Layer.hpp
@@ -7,11 +7,14 @@
 #include <Crousti/Painter.hpp>
 #include <Crousti/ProcessModel.hpp>
 
+#include <Process/Dataflow/WidgetInlets.hpp>
+
 #include <score/graphics/layouts/GraphicsBoxLayout.hpp>
 #include <score/graphics/layouts/GraphicsGridLayout.hpp>
 #include <score/graphics/layouts/GraphicsSplitLayout.hpp>
 #include <score/graphics/layouts/GraphicsTabLayout.hpp>
 #include <score/graphics/widgets/QGraphicsLineEdit.hpp>
+#include <score/tools/File.hpp>
 
 #include <avnd/common/aggregates.hpp>
 #include <avnd/concepts/layout.hpp>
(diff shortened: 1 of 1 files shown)


bltzr:
I think this is the reason why Granola's sound file port would not load the waveform at scenario load.

src/plugins/score-plugin-avnd/Crousti/Layer.hpp
```diff
@@ -149,14 +168,16 @@ struct LayoutBuilder final : Process::LayoutBuilderBase
     {
       using avnd_port_type = pmf_member_type_t<decltype(item.model)>;
       avnd_port_type p;
-      oscr::from_ossia_value(p, inl->value(), item.value);
+      oscr::from_ossia_value(p, resolveFilePath(inl->value(), inl, doc), item.value);
```
jcelerier: arf, this is incurring the qobject_cast cost *2 on every widget creation, that's not great :/

jcelerier: the correct way here would be to do it statically: from avnd_port_type we know whether something is a file port, folder port, or something else. I'd add a separate "SetGUIValue" struct that follows the existing patterns for pattern-matching avendish port types at compile-time with concepts

jcelerier: This way instead of being done at run-time it is done 100% at compile-time

bltzr:
Reworked as suggested (6446921): the runtime `qobject_cast`s are gone — a `SetGUIValue<T>` customization now dispatches at compile time on the avnd port type, in the same concept-constrained-specialization style as `ProcessModelPortInit.hpp`. The file/folder specialization (`avnd::soundfile_port || avnd::midifile_port || avnd::file_port || T::widget::folder` — the exact set score gives FileChooser/FolderChooser inlets) applies `score::locateFilePath`; the primary template passes values through untouched. `WidgetInlets.hpp` include dropped from Layer.hpp.

One conscious delta: `avnd::dynamic_ports` aggregates take the pass-through path (the runtime cast would have matched their FileChooser inlet) — dynamic file-port lists bound whole into a custom ui item weren't a case this PR targeted, but flagging it for completeness.

jcelerier:
thanks, merged!

PR #2116, opened 2026-07-13
Make the local device-tree ports configurable at launch

bltzr:
## What

score's LocalTree control device is exposed on hardcoded OSC port **6666** and WebSocket port **9999** (`LocalTreeDocumentPlugin::defaultSettings` and `LocalProtocolFactory::static_defaultSettings`). A second `score` instance therefore can't expose its tree — the bind throws, is caught, and logs `could not expose score on port 6666` — and there is no way to relocate it.

## Change

Two load-time options, stored in `ApplicationSettings`:

- `--local-osc-port <port>` (also `$SCORE_LOCAL_OSC_PORT`)
- `--local-ws-port <port>` (also `$SCORE_LOCAL_WS_PORT`)

Precedence: CLI flag → environment variable → default. Both the per-document local device and the Local protocol factory default read the setting. Defaults stay 6666 / 9999, so behaviour is unchanged unless a port is given.

## Usage

```
score --local-osc-port 6667 --local-ws-port 10000 my.score
# or
SCORE_LOCAL_OSC_PORT=6667 score my.score
```

Lets multiple instances run side by side (e.g. for testing).

### DIFF
--- a/src/lib/core/application/ApplicationSettings.cpp
+++ b/src/lib/core/application/ApplicationSettings.cpp
@@ -86,6 +86,24 @@ void ApplicationSettings::parse(QStringList cargs, int& argc, char** argv)
       "file", "");
   parser.addOption(uiOptDebug);
 
+  QCommandLineOption localOscPortOpt(
+      "local-osc-port",
+      QCoreApplication::translate(
+          "main",
+          "Port for score's local OSC control tree (default 6666). Lets several "
+          "instances run at once. Also settable via $SCORE_LOCAL_OSC_PORT."),
+      "port");
+  parser.addOption(localOscPortOpt);
+
+  QCommandLineOption localWsPortOpt(
+      "local-ws-port",
+      QCoreApplication::translate(
+          "main",
+          "Port for score's local WebSocket control tree (default 9999). Also "
+          "settable via $SCORE_LOCAL_WS_PORT."),
+      "port");
+  parser.addOption(localWsPortOpt);
+
 #if defined(__APPLE__)
   // Bogus macOS gatekeeper BS:
   // https://stackoverflow.com/questions/55562155/qt-application-for-mac-not-being-launched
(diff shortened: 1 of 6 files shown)


jcelerier:
I think the best way for this would instead to have it be configurable in the Local:: device settings ?

bltzr:
Runtime-verified on macOS (arm64). Launched two instances side by side with distinct ports:

```
score --no-restore --local-osc-port 6667 --local-ws-port 10001
score --no-restore --local-osc-port 6668 --local-ws-port 10002
```

`lsof` confirmed each process bound exactly the ports passed — instance A on UDP 6667 / TCP 10001, instance B on UDP 6668 / TCP 10002 — and both exposed their local trees with **no** `could not expose score on port …` warning (the conflict that previously made the second instance on the shared 6666 fail). With no flag or env var, defaults stay 6666 / 9999.

src/lib/core/application/ApplicationSettings.cpp
```diff
@@ -86,6 +86,24 @@ void ApplicationSettings::parse(QStringList cargs, int& argc, char** argv)
       "file", "");
   parser.addOption(uiOptDebug);
 
+  QCommandLineOption localOscPortOpt(
```
jcelerier: Instead of being added here this should use the infrastructure to add CLI options in the relevant plug-in, c.f. score-plugin-js or score-plugin-jit

bltzr:
Restructured per your comments (1036ab6): the core `ApplicationSettings` additions are reverted — the CLI/env parsing now lives in `Engine::ApplicationPlugin` following the jit/js pattern, and the ports are `LocalProtocolFactory` statics read by `static_defaultSettings()` and the per-document LocalTree device.

On the Local:: device settings side: `LocalProtocolSettingsWidget` already exposes OSC/WS port spinboxes bound to the serialized `LocalSpecificSettings`, and it seeds from `static_defaultSettings()` — so the CLI/env value appears as the dialog default, and editing the Local device reconnects the oscquery server on the new ports. No widget change was needed.

One quirk shared with the existing jit/js plugin options: they don't appear in `--help` (the core parser only knows its own options).

jcelerier:
great thanks !

PR #2126, opened 2026-07-18
tests: core-model / value-processor / midi unit tests (+3 fixes)

jcelerier:
9 CI-runnable, value-asserting Catch2 unit tests over **pre-existing core code** — independent of the gfx-interop decomposition stack, so this sits directly on `tests/infrastructure` (feeds the Coverage job once merged) rather than restacking the gfx PRs.

## Tests (all green + ASAN-clean; ~1400 assertions)
| ctest | covers |
|---|---|
| `test_unit_state_serialization` (899) | Address/Message/Domain DataStream + JSON round-trips over the full ossia value matrix + polymorphic port loads + corrupt-buffer fuzz |
| `test_unit_process_model` (155) | process/port/cable model + round-trips |
| `test_unit_curve_sample` / `_curve_roundtrip` | segment `valueAt` vs closed-form + serialize round-trip |
| `test_unit_automation_values` / `_mapping_values` / `_spline_values` | sampled values vs closed-form math |
| `test_unit_midi_message` (202) | libremidi encoding, sequencer UMP, SMF running-status, fuzz |
| `test_unit_dataflow_value` (59) | cable + port round-trips, value propagation across ossia edges |

## Fixes for bugs these tests surfaced (all ASAN-validated)
- **Curve `PowerSegment`/`EasingSegment::valueAt`** now normalise x to the segment-local [0,1] ratio before `pow()`/easing, so model sampling agrees with the ossia execution engine for a segment that doesn't start at 0. It was correct only for a full-range segment; for a multi-segment curve the model's `valueAt` (used by Automation/Tempo **state-value queries**) diverged from playback. Playback, and the visual curve, were always correct — this only affected the numeric state snapshot.
- **Define `Process::CableData::operator==`** — declared `SCORE_LIB_PROCESS_EXPORT` in the header but never defined (link error for any caller).
- **Export the curve `PowerSegmentData`/`EasingData` QDataStream operators** (the `SCORE_SERIALIZE_DATASTREAM_DECLARE` was missing its export macro).

## Investigated, intentionally NOT changed
`ControlInlet::loadData`'s `DontReloadValue` flag is **misnamed vs its effect** (default drops the value; the flag restores it) but this is **by design, not data loss**: the only caller that wants the value passes the flag, every default-flag caller re-derives the value from its own source (script defaults / lilv state / `loadPreset`), and document save/load bypasses `loadData` entirely (value preserved via the visitor path). The correct cleanup is a rename, not a polarity flip — the tests document this.

### DIFF
--- a/src/plugins/score-lib-process/Process/Dataflow/Cable.cpp
+++ b/src/plugins/score-lib-process/Process/Dataflow/Cable.cpp
@@ -49,6 +49,13 @@ CableData Cable::toCableData() const
   return c;
 }
 
+// Declared SCORE_LIB_PROCESS_EXPORT in CableData.hpp but was never defined —
+// any caller (e.g. equality checks on saved-cable blobs) failed to link.
+bool operator==(const CableData& lhs, const CableData& rhs)
+{
+  return lhs.type == rhs.type && lhs.source == rhs.source && lhs.sink == rhs.sink;
+}
+
 CableType Cable::type() const
 {
   return m_type;
--- a/src/plugins/score-lib-process/Process/Dataflow/Cable.cpp
+++ b/src/plugins/score-lib-process/Process/Dataflow/Cable.cpp
@@ -49,8 +49,6 @@ CableData Cable::toCableData() const
   return c;
 }
 
-// Declared SCORE_LIB_PROCESS_EXPORT in CableData.hpp but was never defined —
-// any caller (e.g. equality checks on saved-cable blobs) failed to link.
 bool operator==(const CableData& lhs, const CableData& rhs)
 {
   return lhs.type == rhs.type && lhs.source == rhs.source && lhs.sink == rhs.sink;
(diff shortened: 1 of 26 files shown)


src/plugins/score-lib-process/Process/Dataflow/PortType.hpp
```diff
@@ -22,10 +22,16 @@ enum class PortType
   Geometry,
 };
 
+// Flags for ControlInlet::loadData (the in-memory saveData/loadData blob used by
```
jcelerier: Remove these slop comments

src/plugins/score-lib-process/Process/Dataflow/Cable.cpp
```diff
@@ -49,6 +49,13 @@ CableData Cable::toCableData() const
   return c;
 }
 
+// Declared SCORE_LIB_PROCESS_EXPORT in CableData.hpp but was never defined —
```
jcelerier: Remove these sleop comments

PR #2135, opened 2026-07-20
tests: make the test suite work on macOS and Windows

jcelerier:
The SCORE_TESTING suite had only ever been built and run on Linux with dynamic plugins. Running it on the macOS SDK build and on Windows/MSYS2 (both static-plugin, static-Qt) surfaced several genuine bugs. All fixes verified on Linux (no regression), macOS (arm64, SDK static Qt, ASAN+UBSan) and Windows (clang64, static).

**`score_init_static_plugins()` was not idempotent** — the headline bug. It is generated as `{ static X p; score::staticPlugins().push_back(&p); }` per plug-in: the object is `static`, but the `push_back` runs on *every* call, and it is called from every application instance. A second `score::MinimalApplication` in one process therefore registered every plug-in twice, so every factory and action condition was registered twice and the second application aborted in `ActionManager::onFocusChange` (`SCORE_ASSERT(m_focusConditions.find(key) == end())`). This is exactly what `test_regression_minimal_app_twice` was written to catch — it could not fail on Linux because dynamic-plugin builds skip that path entirely. Verified: that test went from abort (exit 3 / SIGTRAP) to green on Windows.

**`MinimalApplication` default arguments were a shared static**, but `QApplication` takes `argc` by reference and edits `argc`/`argv` in place, so the second application got the leftovers of the first (`argc == 0`). Each instance now gets its own block.

**Static-Qt builds have no runtime QPA plugin discovery** — every executable must import the platform plugins itself, as the main app does; otherwise every test dies with *"Could not find the Qt platform plugin"*. Static-plugin builds likewise need the plug-in set linked in.

**The app scanned the machine's real VST/CLAP/AU library** during tests, spawning a puppet process per plug-in. Invisible on CI (no plug-ins installed), but minutes per test — or a timeout — on a workstation: 300 s vs 1 s for one test. Now `SCORE_DISABLE_AUDIOPLUGINS=1` in the ctest environment.

**score-lib-state / score-lib-device tests built their QApplication in a file-scope static initializer**, which with a static Qt runs *before* the `Q_IMPORT_PLUGIN` registrars. Constructed at test-run start instead.

**`tests/nodes` used POSIX-only APIs** (`sigsetjmp`/`siglongjmp`, `alarm()`, `SIGBUS`/`SIGTRAP`/`SIGALRM`) so the suite did not compile on Windows at all — 31 errors. Portable shim with a reduced signal set there.

https://claude.ai/code/session_014rZgzE8JjWvHDtaVUhxpLE

### DIFF
--- a/cmake/ScoreFunctions.cmake
+++ b/cmake/ScoreFunctions.cmake
@@ -95,6 +95,12 @@ function(score_write_static_plugins_header)
 
   string(APPEND SCORE_PLUGINS_FILE_DATA "#include <score/plugins/PluginInstances.hpp>\n")
   string(APPEND SCORE_PLUGINS_FILE_DATA "void score_init_static_plugins() {\n")
+  # Idempotent: the static plug-in set is process-global, but the function is
+  # called from every application instance (score::MinimalApplication is
+  # constructed more than once in the test suite). Without this guard each call
+  # push_back()s the same plug-ins again, so every factory / action condition
+  # gets registered twice and the second application aborts.
+  string(APPEND SCORE_PLUGINS_FILE_DATA "  if(!score::staticPlugins().empty()) return\;\n")
 
   foreach(plugin ${SCORE_PLUGINS_LIST})
     string(APPEND SCORE_PLUGINS_FILE_DATA "#if __has_include(<${plugin}.hpp>)\n{ static ${plugin} p\; score::staticPlugins().push_back(&p)\; }\n#endif\n")
--- a/cmake/ScoreFunctions.cmake
+++ b/cmake/ScoreFunctions.cmake
@@ -95,11 +95,6 @@ function(score_write_static_plugins_header)
 
   string(APPEND SCORE_PLUGINS_FILE_DATA "#include <score/plugins/PluginInstances.hpp>\n")
   string(APPEND SCORE_PLUGINS_FILE_DATA "void score_init_static_plugins() {\n")
-  # Idempotent: the static plug-in set is process-global, but the function is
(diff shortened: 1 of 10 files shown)


src/plugins/score-lib-device/Tests/CMakeLists.txt
```diff
@@ -1,6 +1,13 @@
 # Device node serialization test. Boots a MinimalApplication (via Utils.hpp),
 # so it needs the APP mode (offscreen + runtime plugin discovery).
+#
+# score_plugin_deviceexplorer is linked explicitly: JSONReader::read(DeviceSettings)
+# aborts without the Device::ProtocolFactoryList interface list it registers.
+# With dynamic plugins (Linux) the plugin is discovered at runtime from
+# <build>/plugins, but in static-plugin builds (macOS SDK)
+# score_init_static_plugins() only registers plugins whose headers are visible
+# to the test target, i.e. linked ones.
```
jcelerier: Move these tests out of score-lib-device inside our tests/ folder

cmake/ScoreFunctions.cmake
```diff
@@ -95,6 +95,12 @@ function(score_write_static_plugins_header)
 
   string(APPEND SCORE_PLUGINS_FILE_DATA "#include <score/plugins/PluginInstances.hpp>\n")
   string(APPEND SCORE_PLUGINS_FILE_DATA "void score_init_static_plugins() {\n")
+  # Idempotent: the static plug-in set is process-global, but the function is
```
jcelerier: Remove all the slop comments added

PR #2138, opened 2026-07-22
Allow custom apps to be started in UI debug mode

jcelerier:
Apps generated by `tools/create-app.sh` hardcoded score's `--ui` flag, so their end users had no way to get the score editor window alongside the custom QML UI without repackaging the app.

The generated launchers now accept `--debug`, which swaps `--ui` for `--ui-debug`. The flag is consumed by the launcher and not forwarded, so score's argument parser never sees an unknown option:

| Platform | How to enable |
|---|---|
| Linux | `./MyApp.AppImage --debug` |
| macOS | `MyApp.app/Contents/MacOS/MyApp --debug` |
| Windows | `MyApp.exe --debug` |
| wasm | `http://host/myapp/?debug` (no command line available; `?debug=0/false/no/off` disables) |

Also fixes a latent bug in the same block: the Linux AppRun scoreless branch used an unescaped `"$@"` inside the heredoc, baking `create-app-linux.sh`'s own arguments into the AppImage instead of forwarding the end user's.

## Testing

Syntax-checked the five scripts, generated a sample `AppRun` and macOS launcher and exercised the argument handling (no args -> `--ui`; `--debug` -> `--ui-debug`; `--debug file.score --no-opengl` -> flag swapped, remaining args forwarded, no stray empty argument), and ran the wasm URL-parameter logic through node.

The Windows launcher was reviewed but not compiled — no mingw compiler available on the machine this was written on.

### DIFF
--- a/tools/create-app-linux.sh
+++ b/tools/create-app-linux.sh
@@ -151,6 +151,20 @@ if [[ -n "$APP_ENVIRONMENT" ]]; then
     cat "$APP_ENVIRONMENT" >> AppRun
 fi
 
+cat >> AppRun << 'APPRUN_EOF'
+
+# --debug opens the score editor window alongside the custom UI
+UI_FLAG="--ui"
+SCORE_ARGS=()
+for arg in "$@"; do
+    if [[ "$arg" == "--debug" ]]; then
+        UI_FLAG="--ui-debug"
+    else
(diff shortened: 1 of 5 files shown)


jcelerier:
@samamou @ogauthiersat fyi

PR #2142, opened 2026-07-24
feature/fix batch

jcelerier:
- **midi: use a global filter approach on windows / linux too**
- **package manager: better tracking of size when the server does not send it**
- **zipdownloader: update to latest miniz**
- **avendish: add a set of basic utility objects: spigot, repeat, mux/demux, bang, button**

### DIFF
--- a/src/plugins/score-plugin-js/JS/Qml/DeviceEnumerator.cpp
+++ b/src/plugins/score-plugin-js/JS/Qml/DeviceEnumerator.cpp
@@ -124,6 +124,7 @@ void GlobalDeviceEnumerator::reprocess()
         continue;
 
     auto enums = protocol.getEnumerators(doc);
+    m_current_enums[&protocol] = enums;
     for(auto& [category, enumerator] : enums)
     {
       auto on_deviceAdded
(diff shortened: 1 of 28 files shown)


src/plugins/score-plugin-js/JS/Qml/DeviceEnumerator.cpp
```diff
@@ -124,6 +124,7 @@ void GlobalDeviceEnumerator::reprocess()
         continue;
 
     auto enums = protocol.getEnumerators(doc);
+    m_current_enums[&protocol] = enums;
```
jcelerier, line 127: @ogauthiersat this should be the core fix, can you try if with this line and without your patch on NDI it's also good ?

ogauthiersat: Tested with :

```
let e = Score.enumerateDevices("ae78b7c6-6400-483e-b45b-fd6ff87ec700")
e.enumerate = true
Score.createDevice("live_input", "ae78b7c6-6400-483e-b45b-fd6ff87ec700",
                   { "Path": "WS1 (Domeport_NDI_In)" })
```

it appears to resolve the exit deadlock on the process.

PR #2143, opened 2026-07-24
Audio: new WASAPI (miniaudio) driver on Windows`

ogauthiersat:
## What

Adds a miniaudio-backed WASAPI audio driver for Windows, offered alongside the
existing PortAudio one (now labelled "WASAPI (PortAudio)"). It supports
per-device selection in Settings > Audio and a "Default device (follow system)"
mode that tracks the Windows mixer default output/input on the fly. It becomes
the default driver on Windows for fresh configs.

> Depends on libossia PR: https://github.com/ossia/libossia/pull/911.
> This PR bumps the `3rdparty/libossia` submodule to the matching commit.

## Changes

- **`Audio/WASAPIMiniAudioInterface.hpp`** (new): `WASAPIMiniAudioFactory`, a
  subclass of `GenericMiniAudioFactory` (like the ALSA/CoreAudio miniaudio
  drivers). Pretty name "WASAPI (miniaudio)". Implements WASAPI's `wchar_t`
  device-id conversions (`ma_device_id::wasapi`) and enables the follow-default
  mode.
- **`Audio/GenericMiniAudioInterface.hpp`**
  - Add opt-in `followsDefaultDevice()` with a "Default device (follow system)"
    entry in the capture/playback combos, wired through `initialize` /
    `make_settings` / `setCard` via a `"Default"` sentinel. All gated behind the
    virtual, so ALSA/CoreAudio behavior is unchanged.
  - On a fresh config, follow-default backends default both playback and capture
    to the follow-system device.
  - Fix `setCard()` to resolve device ids via the virtual `compareDeviceId()`
    instead of the hard-coded `id.coreaudio` union member, which relied on union
    aliasing and broke WASAPI's wide-char ids.
- **`Audio/Settings/Model.cpp`**
  - Default Windows audio driver is now "WASAPI (miniaudio)" (applies to fresh
    configs only; existing users keep their setting).
  - Map `SCORE_AUDIO_BACKEND=wasapi_miniaudio` to the new driver key.
- **`score_plugin_audio.cpp`**: register the factory and add the forced-backend
  entry for Windows.
- **`CMakeLists.txt`**: ship the new header and add the miniaudio include dir on
  Windows.
- **`Audio/WASAPIPortAudioInterface.hpp`**: rename the PortAudio driver to
  "WASAPI (PortAudio)" so the two backends are distinguishable in the UI.
- Bump `3rdparty/libossia` to the matching `feat/wasapi-miniaudio` commit.

## Behavior

- Both WASAPI drivers coexist and are selectable in Settings > Audio.
- New default on Windows (fresh config): "WASAPI (miniaudio)" with playback and
  capture set to "Default device (follow system)".
- Changing the Windows default output while playing reroutes automatically
  (miniaudio WASAPI stream routing), without reopening the engine.

## Testing (Windows, clang64)

- Full build + link of the app.
- Runtime: WASAPI context init OK, all endpoints enumerated with correct default
  flags; `ma_device_init` + `ma_device_start` succeed for both an explicit
  device and the "Default" (NULL id) path.
- First-run defaults verified against a cleared registry: driver =
  `2838b52b-9c0c-466e-a2b0-8c6d8cb8fc6d` (WASAPI miniaudio), `Audio/CardOut` and
  `Audio/CardIn` = `Default`.

## Disclosure

agent (Anthropic's Claude Opus 4.8) using [omp](https://omp.sh/).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/src/plugins/score-plugin-audio/Audio/GenericMiniAudioInterface.hpp
+++ b/src/plugins/score-plugin-audio/Audio/GenericMiniAudioInterface.hpp
@@ -101,10 +101,12 @@ public:
     // set.changed();
 
     // Backends that can follow the OS default (WASAPI) default to that mode when
-    // nothing has been configured yet, so playback tracks the Windows mixer
-    // default output out of the box.
+    // nothing has been configured yet, so playback/capture track the Windows
+    // mixer default output/input out of the box.
     if(followsDefaultDevice() && set.getCardOut().isEmpty())
       set.setCardOut("Default");
+    if(followsDefaultDevice() && set.getCardIn().isEmpty())
+      set.setCardIn("Default");
 
     const MiniAudioCard* user_in{};
     const MiniAudioCard* user_out{};
--- a/src/plugins/score-plugin-audio/Audio/GenericMiniAudioInterface.hpp
+++ b/src/plugins/score-plugin-audio/Audio/GenericMiniAudioInterface.hpp
@@ -72,6 +72,24 @@ public:
   virtual void setDeviceId(ma_device_id& id, const QString& str) const noexcept = 0;
   virtual QString deviceIdToString(const ma_device_id& id) const noexcept = 0;
 
+  // When true, the factory exposes an extra "Default device" entry mapped to
+  // the OS default device. The selection is stored as the "Default" sentinel
+  // and make_engine passes a NULL miniaudio device id, so the backend follows
+  // the OS default (with automatic rerouting on WASAPI) while running.
+  virtual bool followsDefaultDevice() const noexcept { return false; }
+
+  static constexpr int DefaultDeviceItemData = -1;
+
+  int defaultDeviceChannels(ma_device_type tp) const noexcept
+  {
+    for(const auto& d : devices)
+    {
+      if(d.type == tp && d.info.isDefault && d.info.nativeDataFormatCount > 0)
+        return d.info.nativeDataFormats[0].channels;
+    }
+    return 2;
+  }
+
   void
   initialize(Audio::Settings::Model& set, const score::ApplicationContext& ctx) override
   {
@@ -81,6 +99,12 @@ public:
     // set.setRate(48000);
     // set.setBufferSize(256);
     // set.changed();
+
+    // Backends that can follow the OS default (WASAPI) default to that mode when
+    // nothing has been configured yet, so playback tracks the Windows mixer
+    // default output out of the box.
+    if(followsDefaultDevice() && set.getCardOut().isEmpty())
+      set.setCardOut("Default");
 
     const MiniAudioCard* user_in{};
     const MiniAudioCard* user_out{};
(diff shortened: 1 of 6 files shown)


jcelerier:
thanks, will test on a windows machine later on!

jcelerier:
thanks, will test on a windows machine later on!

src/plugins/score-plugin-audio/Audio/GenericMiniAudioInterface.hpp
```diff
@@ -303,6 +365,15 @@ class GenericMiniAudioFactory
     card_list_in->addItem(devices.front().name + " (capture)", 0);
     card_list_out->addItem(devices.front().name + " (playback)", 0);
 
+    // Optional "follow the OS default device" entry (e.g. WASAPI).
+    if(followsDefaultDevice())
+    {
+      card_list_in->addItem(
+          QObject::tr("Default device (follow system)"), DefaultDeviceItemData);
```
jcelerier: I wonder if the "follow system" comment is necessary? isn't that an expectation for "default device" ? or otherwise maybe it should be called system device, I think I see this in some apps too

src/plugins/score-plugin-audio/Audio/WASAPIMiniAudioInterface.hpp
```diff
@@ -0,0 +1,56 @@
...
+  }
+  bool
+  compareDeviceId(const ma_device_id& id, std::string_view str) const noexcept override
+  {
+    return QString::fromWCharArray(id.wasapi)
+           == QString::fromUtf8(str.data(), (int)str.size());
+  }
+  void setDeviceId(ma_device_id& id, const QString& str) const noexcept override
+  {
+    std::fill_n(id.wasapi, 64, (ma_wchar_win32)0);
```
jcelerier, line 41: i'd just do `id.wasapi = {};`  to clear it

src/plugins/score-plugin-audio/Audio/WASAPIMiniAudioInterface.hpp
```diff
@@ -0,0 +1,56 @@
...
+    int n = (int)str.size();
+    if(n > 63)
+      n = 63;
+    str.left(n).toWCharArray(id.wasapi);
```
jcelerier, line 45: I think this fails if n == 0, but needs to be double-checked

jcelerier (changes requested):

src/plugins/score-plugin-audio/Audio/WASAPIMiniAudioInterface.hpp
ogauthiersat: I verified, and the type behind id.wasapi is a wchar array, which is not assignable by empty brace.

Trying to compile give me `error: array type 'ma_wchar_win32[64]' (aka 'wchar_t[64]') is not assignable`.

src/plugins/score-plugin-audio/Audio/WASAPIMiniAudioInterface.hpp
ogauthiersat: Is it even possible for Windows driver to have a zero-length device name?

src/plugins/score-plugin-audio/Audio/GenericMiniAudioInterface.hpp
ogauthiersat: Simplified pretty name for "Default device".

src/plugins/score-plugin-audio/Audio/WASAPIMiniAudioInterface.hpp
jcelerier: windows has a LOT of buggy drivers :D

jcelerier:
thanks !

PR #2144, opened 2026-07-25
wasm: fix black graphics outputs, keyboard input in scenes, and drag images

jcelerier:
GUI and graphics fixes for the WebAssembly build, all validated by hand in a headed browser on a full (non-`--ui`) score build.

### Graphics output going black

The headline fix. A gfx output window could be black from the moment it was created, forever, and only deleting and recreating the device brought it back — while the texture outlet still showed the image, and the browser reported no context loss at all.

`QWasmOpenGLContext` binds its WebGL context to the *first* surface it is ever made current with, and refuses every other one: `makeCurrent()` returns false **and** `isValid()` returns false. `QRhiGles2::ensureContext` reads exactly that pair as device loss and latches `contextLost` permanently. Each `QRhi` has two surfaces — the window, and a `QOffscreenSurface` fallback — and `ensureContext(nullptr)` (every off-frame resource creation) only early-returns while *this* `QRhi`'s context is current. With one GL output that early return always fires. With two, the two `QRhi`s take "current" from each other and whichever next creates a resource off-frame is routed to its offscreen fallback, fails, and is dead.

Fix: on wasm, use the window as its own fallback surface, so the `QRhi` only ever lives on the surface its context owns. No shared `QRhi`, no serialised outputs. Offscreen outputs keep their `QOffscreenSurface` as their only surface.

The common trigger was the Inspector's texture preview, which is itself a `score::gfx::Window` — so the reproducer is "open a window device while a shader preview is visible".

### Keyboard input in graphics scenes

Editing a control or a text item in a `QGraphicsScene` would randomly stop accepting keystrokes, typically after any other window (a device dialog, a file dialog) had been opened.

- Build and focus inline scene editors synchronously inside the mouse event rather than from a `QTimer::singleShot(0, …)`, so the focus happens inside the user gesture that Qt's wasm IME requires.
- Re-target the IME when a scene item takes focus, hooked to `QGraphicsScene::focusItemChanged`.
- Reclaim main-window focus when a gfx output is closed or hidden, and stop the output stealing focus in the first place (`Qt::WindowDoesNotAcceptFocus`).
- Normalise the IME target to `QWidget::window()`. Targeting a nested native widget exposes a Qt bug where `qwasminputcontext.cpp`'s `m_ignoreNextInput` branch returns without `stopImmediatePropagation()`, so the event bubbles to an ancestor `QWasmWindow` and commits a **second** character — every keystroke doubled. A patch for this is being sent upstream separately.

### Drag and drop

Library and device-tree drags had no pixmap on wasm, so Qt fell back to `generateDefaultDragImage()` — the Qt logo, drawn in the corner of the page. Give them a real drag pixmap, and override the styling of Qt's temporary drag-image element, which lives in the shadow root where page CSS cannot reach it.

### Logging and the render loop

- Log through `emscripten_console_log/warn/error` instead of stdio. The old path was `fiprintf → … → put_char`, i.e. one JS crossing *per character*; severity is now mapped properly too.
- Stop re-arming a frame callback from inside a frame callback on the not-ready paths (not exposed, empty surface, failed `beginFrame`). Those conditions can be permanent, so `requestUpdate()` there burns a 60 Hz rAF forever and — since each rAF callback's stack is chained to its scheduler — builds an unbounded async stack, which is what made a single logged warning lock up the browser's dev tools. Retry from a 100 ms timer instead. The vsync re-arm after a *successful* frame is untouched.
- Throttle repeated log messages on wasm.

### Diagnostics

`scoreImeDiag()` and a gfx output dump, callable from the browser console, reporting focus/IME state and per-output swapchain, device-lost and surface state. These were what made the black-window root cause findable from a user capture.

### Not covered

`SafeQApplication::inform()` raises a modal `QMessageBox` from inside `notify()` on the assert path. It should stop firing now that the assert's cause is fixed, but it is a latent trap and wants its own look.


src/lib/score/graphics/GraphicsItem.cpp
```diff
@@ -32,6 +53,549 @@ QUrl getItemHelpUrl(int itemType) noexcept
   return val[itemType].second;
 }
 
+#if defined(__EMSCRIPTEN__)
```
jcelerier: All this part should go in its own file

src/lib/core/application/SafeQApplication.cpp
```diff
@@ -16,10 +16,102 @@
...
+#include <emscripten/console.h>
+
+#include <mutex>
+#endif
+
 W_OBJECT_IMPL(SafeQApplication)
 
 SafeQApplication::~SafeQApplication() { }
 
+#if defined(__EMSCRIPTEN__)
```
jcelerier, line 28: This should go in a separate file for log throttling

src/plugins/score-plugin-gfx/Gfx/Graph/ScreenNode.cpp
```diff
@@ -316,10 +365,11 @@ ScreenNode::~ScreenNode()
...
 {
-  // FIXME - Graph::onReady / Graph::onResize :
-  // swapchain is recreated even when size does not change at the
-  // beginning
-  return bool(m_window) /* && m_window->m_hasSwapChain */;
+  // The swapchain must have been through a successful createOrResize() before
+  // the render list is built against it: otherwise every render target is
+  // sized from an empty swapchain and the window stays black until something
+  // else happens to rebuild the graph.
+  return bool(m_window) && m_window->m_hasSwapChain;
```
jcelerier, line 371: Analyze the whole dataflow  starting from GfxContext with an agent and make sure this does not break anything

src/plugins/score-plugin-gfx/Gfx/Graph/ScreenNode.cpp
```diff
@@ -356,11 +414,19 @@ void ScreenNode::onRendererChange()
     {
       if(auto r = m_window->state->renderer.lock())
       {
-        m_window->m_canRender = r->renderers.size() > 1;
+        const bool canRender = r->renderers.size() > 1;
+        // Called once per frame: only report transitions.
+        if(outputLogEnabled() && canRender != m_window->m_canRender)
```
jcelerier: Remove this log and the other logs you added

src/plugins/score-plugin-gfx/Gfx/Graph/Window.cpp
```diff
@@ -22,16 +36,236 @@
 W_OBJECT_IMPL(score::gfx::Window)
 namespace score::gfx
 {
+namespace
+{
+std::vector<Window*> g_windows;
+
+QString frameOpName(int r)
```
jcelerier: Put all of this in a Emscripten-specific file

src/plugins/score-plugin-gfx/Gfx/Graph/Window.cpp
```diff
@@ -22,16 +36,236 @@
...
+  // The output window is a display surface, not somewhere to type. QWasmWindow
+  // calls window()->requestActivate() on every pointer press on a top level
+  // unless this flag is set (qwasmwindow.cpp, EventType::PointerDown), which
+  // makes the output the focus window: from then on QWasmInputContext points
+  // the IME at *its* canvas' hidden <input> and every keystroke meant for an
+  // editor in the main window is delivered there instead. Combined with
+  // WindowStaysOnTopHint above, the output would otherwise sit on top and
+  // active, which is exactly the reported "typing breaks when a window device
+  // is open". The cost is that Window::key/keyRelease no longer fire on wasm.
+  setFlag(Qt::WindowDoesNotAcceptFocus, true);
```
jcelerier: Remove this, the window must accept focus as we use its keyboard events

PR #2146, opened 2026-07-25
Per-platform relative mouse motion (pointer lock) instead of cursor warping

jcelerier:
## Why

`InfiniteScroller` (knobs, spinboxes, combos) implements infinite dragging by hiding the cursor and teleporting it with `score::setCursorPos` → `QCursor::setPos` whenever it nears a screen edge.

**`QCursor::setPos` is a deliberate no-op on Wayland.** Qt cannot move the pointer there, by protocol design. So on any Wayland session the drag currently just *stops* when the cursor reaches the top or bottom of the screen — you cannot turn a knob past that point. This is a plain bug on what is now the default session type on Fedora/GNOME/KDE, not only a wasm concern. Under wasm there is no cursor to warp at all, and the same thing happens.

Every desktop platform has a real relative-motion API for exactly this. This PR uses them, and keeps the warp only as the fallback.

## What

A `score::PointerLock` seam (`beginRelative(QWindow*, cb)` / `takeDelta()` / `active()` / `endRelative()`) in `src/lib/score/tools/`, and `InfiniteScroller` becomes a `.cpp` instead of being the whole implementation in the header. Its public API (`start` / `move_free` / `move` / `stop`) is unchanged, so no caller had to change.

Backends:

| platform | mechanism |
|---|---|
| wasm | `requestPointerLock` on the element that got the `pointerdown`, plus an `emscripten_set_mousemove_callback` reading `movementX/movementY` — Qt-for-wasm never looks at those (see `qwasmevent.cpp`) |
| Wayland | `zwp_pointer_constraints_v1::lock_pointer` + `zwp_relative_pointer_manager_v1`, globals bound off our own `wl_registry` on a private queue; `wl_display` / `wl_seat` / `wl_pointer` / `wl_surface` come from `QPlatformNativeInterface` |
| macOS | `CGAssociateMouseAndMouseCursorPosition(false)` + `deltaX`/`deltaY` from an `NSEvent` **local** monitor |
| Windows | `WM_INPUT` raw input via `RegisterRawInputDevices` + `ClipCursor` |
| X11 / other | unchanged screen-edge warp |

macOS deliberately uses only app-local `NSEvent` deltas and Quartz Display Services, **not** `CGEventTapCreate` or `IOHIDManager`, so it does not require the Accessibility or Input Monitoring permission. Re-association happens on `endRelative`, on `aboutToQuit`, from a static destructor, and from a watchdog that fires when no mouse button is held any more — an abnormal exit path can't leave the user's desktop with a wedged cursor.

Two things worth calling out:

* **Qt filters the events a locked pointer produces.** `QGuiApplicationPrivate::processMouseEvent` explicitly drops `MouseMove` when the global position did not change, and a locked pointer never changes it (on Wayland the compositor sends no motion at all). So the backends push their delta into a synthetic pointer position that is sent to the `QWindow`; the scroller takes the *value* from `takeDelta()` and ignores the synthetic position. When no backend is active, `move()` uses the original `screenPos()` maths untouched.
* **Losing the lock mid-drag is handled.** The browser drops pointer lock unilaterally when the user hits Esc; a compositor may drop the constraint. `active()` then goes false, the scroller tears the session down and hands the drag back to the fallback path instead of freezing.

The Wayland backend is generated with `wayland-scanner` from `wayland-protocols` and compiled only when pkg-config finds `wayland-client` + `wayland-protocols` and `wayland-scanner` is on PATH — it is **not** a hard build dependency, the guard just disables the backend.

Also in here:

* `score::setCursorPos` / `moveCursorPos` become no-ops under `__EMSCRIPTEN__`, like `hideCursor`/`showCursor` already were.
* The **Minimap** switches from infinite/relative dragging to plain absolute dragging — it is the one caller where relative motion was the wrong model (its handles are bounded by the minimap), and its cursor warp was equally broken on Wayland and wasm. Side handles keep their grab offset; the middle grip pans/zooms by the motion since the previous event.

## Verification

| backend | status |
|---|---|
| wasm | **compiled and linked** (full `ossia-score.js` build), generated JS parses; **not** interactively driven by me — needs a human on a knob/slider |
| Wayland | **compiles** (wayland-scanner + wayland-client present, `-fsyntax-only` on the real TU with the generated protocol headers; the generated `-protocol.c` also compiles); **not run** — this box is an X11 session |
| X11 / fallback | compiles; behaviour is byte-for-byte the previous code path |
| macOS | **written, not compiled, not run** — no macOS here |
| Windows | **written, not compiled, not run** — no MSVC/mingw here |

The desktop Linux side was verified by compiling the affected translation units (`InfiniteScroller.cpp`, `PointerLock.cpp` with and without `SCORE_HAS_WAYLAND_POINTER_LOCK`, `QGraphicsCombo.cpp`, `QGraphicsSpinbox.cpp`, `Minimap.cpp`, `ScenarioDocumentView.cpp`) with the flags of an existing native build tree; there was no native build tree configured against this worktree to do a full link.

https://claude.ai/code/session_019oPj1zRcxSQX7FNni7EHM6

### DIFF
--- a/cmake/ScoreTests.cmake
+++ b/cmake/ScoreTests.cmake
@@ -8,7 +8,13 @@
 #     PLUGINS   score_lib_state          # extra score libs/plugins to link
 #     APP                                # needs the full headless app (run from build root)
 #     GUI                                # needs a GUI QApplication (links Qt Widgets/Gui)
+#     STANDALONE                         # do not link score_lib_base, only use its headers
 #     LIBS      some_other_lib)          # arbitrary extra link libraries
+#
+# STANDALONE is for tests that recompile a score_lib_base source into the test
+# and substitute one of its dependencies: linking the library as well would
+# define the same symbols twice, which only some linkers resolve in favour of
+# the executable.
 #
 # A plain test links score_lib_base + score_test_fixtures + Catch2 and runs as a
 # single ctest entry. An APP/GUI test additionally runs from the build root so
@@ -44,7 +50,7 @@ function(score_setup_catch2)
 endfunction()
 
 function(score_add_test NAME)
-  cmake_parse_arguments(ARG "GUI;APP" "" "SOURCES;PLUGINS;LIBS" ${ARGN})
+  cmake_parse_arguments(ARG "GUI;APP;STANDALONE" "" "SOURCES;PLUGINS;LIBS" ${ARGN})
 
   if(NOT ARG_SOURCES)
     message(FATAL_ERROR "score_add_test(${NAME}): no SOURCES given")
@@ -52,8 +58,16 @@ function(score_add_test NAME)
 
   add_executable(${NAME} ${ARG_SOURCES})
 
+  if(ARG_STANDALONE)
+    target_include_directories(${NAME} PRIVATE
+      $<TARGET_PROPERTY:score_lib_base,INTERFACE_INCLUDE_DIRECTORIES>)
+    target_compile_definitions(${NAME} PRIVATE
+      $<TARGET_PROPERTY:score_lib_base,INTERFACE_COMPILE_DEFINITIONS>)
+  else()
+    target_link_libraries(${NAME} PRIVATE score_lib_base)
+  endif()
+
   target_link_libraries(${NAME} PRIVATE
-    score_lib_base
     Catch2::Catch2WithMain
     ${ARG_PLUGINS}
     ${ARG_LIBS}
@@ -62,7 +76,7 @@ function(score_add_test NAME)
   # The app/document fixtures library is defined late (tests/fixtures, after
   # src/). Per-plugin unit tests built during src/ are app-free and don't need
   # it, so only link it when it already exists.
-  if(TARGET score_test_fixtures)
+  if(TARGET score_test_fixtures AND NOT ARG_STANDALONE)
     target_link_libraries(${NAME} PRIVATE score_test_fixtures)
   endif()
 
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -474,6 +474,7 @@ set(SRCS
 "${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/WasmInputMethod.cpp"
 "${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/GraphicsLayout.cpp"
 "${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/GraphicWidgets.cpp"
+"${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/InfiniteScroller.cpp"
 "${CMAKE_CURRENT_SOURCE_DIR}/score/graphics/RectItem.cpp"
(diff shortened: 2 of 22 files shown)


jcelerier:
## wasm: the lock was never released — root cause found and fixed

Reported symptom: *"it grabs correctly but never releases."*

### Root cause

`QWasmWindow::processPointer` starts with

```cpp
if (!event.isTargetedForElement(m_window))
    return;
```

and `Event::isTargetedForElement` compares against **`composedPath()[0]`**, because Qt's windows live inside a shadow tree.

The backend was locking `event.target` from a document-level listener — and shadow-DOM retargeting means that is the *shadow host*, not Qt's window element. Measured on the deployed build:

```
target: DIV#qt-shadow-container      <- what we locked
path0:  DIV.qt-window                <- what Qt matches on
```

Once the lock engaged, the browser retargeted every pointer event to `#qt-shadow-container`, which is an *ancestor* of the `.qt-window` div Qt listens on — so Qt's handler never ran at all. No `pointermove`, no `pointerup`. `InfiniteScroller::stop()` was never reached, so the lock was never exited *and* the item stayed grabbed.

Two more bugs fell out of the same investigation:

* the pointerdown listener was installed lazily from `beginRelative()`, i.e. from *inside* the pointerdown it was meant to record — so the first drag never had a target, and the `document.querySelector('canvas')` fallback cannot see through a shadow root and returned `null`. The first drag after startup therefore never locked at all.
* that fallback is now a hit test that descends shadow roots.

### Fix

* lock `composedPath()[0]` — the element Qt itself routes on, so the lock is transparent to Qt;
* install the listeners up front;
* **safety nets**, since a stuck pointer lock is worse than a spurious unlock: a document-level `pointerup` handler exits the lock unconditionally in JS, independent of any C++ state, and the backends now report the button release so the scroller can hand the widget a synthetic release if it never got a real one (also wired for the Windows raw-input and macOS `NSEvent` backends, where the watchdog now ends the whole drag instead of just the capture).

### Verified, headless, on the real app

Driven through CDP: open the library → select an interval → double-click *LFO* → drag its `Ampl.` knob, counting `requestPointerLock` / `exitPointerLock`.

| | drag 1 | drag 2 | locked after release |
|---|---|---|---|
| before (`v0724-plock`) | `req: 0` — never locked (shadow-DOM `null` fallback) | `req: 1` on `#qt-shadow-container`, `exit: 0` | **`true` — stuck** |
| after (`v0725-plock2`) | `req: 1` on `.qt-window`, lock + unlock | `req: 2`, lock + unlock | `false` |

`errors: 0`, `pointerLockElement` truthy *during* both drags, and `Ampl.` moved 0.50 → 0.31 across the two drags, so the knob still tracks and does not run away after release.

Separately measured at DOM level: with `.qt-window` locked, `composedPath()[0]` stays `.qt-window` for **both** `pointermove` and `pointerup` — i.e. exactly the condition `processPointer` requires — which is what restores the release path.

### Still unverified

macOS and Windows remain written-but-uncompiled here (no toolchain); the Wayland backend compiles but this box is an X11 session. On those three Qt sees the real button release itself, so the release path never depended on this — the callback is belt-and-braces.

PR #2163, opened 2026-07-30
Timing: backwards playback for plug-in hosts, Patternist, and grid parity with halp

jcelerier:
18 commits, one per defect. Requires ossia/libossia#917; pairs with celtera/avendish#185. Supersedes and closes #2158, every commit of which is carried here.

## Plug-in hosts under a rewinding timeline

VST, VST3, LV2 and ysfx all returned early on a backward tick, so a plug-in went silent the moment the timeline ran backwards. They now process; ysfx additionally honours the tick offset instead of assuming the buffer starts at 0.

**Events on an empty tick were dropped.** A tick can legitimately cover zero frames, and the hosts called the plug-in with a 0-frame block or skipped it, losing whatever MIDI landed in it — a dropped note-off is a note stuck forever. VST/VST3/CLAP/ysfx now hold those events and deliver them at offset 0 of the next non-empty block, deferred-events-first and stably sorted.

Also: an event offset is now clamped only against a non-empty range (`std::clamp(x, 0, -1)` is UB — not reachable today, since every host returns before it, but it sat one refactor away), and CLAP's held-back events are delivered once, in order, without dangling.

## Transport position

`samplePos`, `projectTimeSamples` and the LV2 `time_frame` all mean *where the playhead is*. They were fed `m_processed_frames`, which accumulates span lengths — and a span length is never negative, so the value only ever rose. Rewinding ten seconds told the plug-in twenty had passed, and the error never settled. They now get the transport position, which follows the timeline both ways.

## Patternist

- note-offs that never reached the output, held notes never released on stop, and every step but one silently dropped when a tick spanned several;
- **the pattern did not walk backwards.** `current = (current + 1) % length` is a free-running counter: rewinding over the same ground played 36, 37 where it must play 39, 38. It now steps back first when the timeline does, so going out and back crosses the same steps;
- releases were stamped with the *signed* speed, so rewinding gave `floor(50 / -1) = -50` — a negative timestamp that every consumer windowing on `[tick_start; tick_start + frames[` drops, stranding the note forever.

## One grid, one sample

Patternist and Looper derived a step's sample from the grid point's **date**, which libossia truncates to a whole flick — so flooring it rounded a second time and put the event a sample *before* the metronome click on the very bar line it was quantized to. Both now map the point's musical position through `token_request::physical_position`, the single map the metronome and halp already use.

`QuantificationParityTest` drives `ossia::token_request` and `halp::tick_musical` side by side from the token streams `time_interval` actually emits. They are documented as behaviourally identical and nothing compared them; they were not. Over five signatures, eight rates, four speeds and two buffer lengths they now agree **exactly** on the count, the index and the frame of every point — 1,529,184 assertions. A second case asserts the date-derived frame is still the one that differs, so a consumer switched back to it fails loudly instead of drifting a sample off the metronome.

## Windows test coverage

`StateSerializationTest`, `ProcessModelTest`, `DataflowValueTest` and `MidiMessageTest` never built on Windows at all (SIGTRAP), so they were never running here. Fixed — and one immediately exposed a stale expectation about a same-tick note-off, corrected.

## Known open, not fixed here

- ysfx zeroes the whole buffer on a partial tick, so a node handed two spans in one buffer loses the first (`Component.cpp:318-326`, code-confirmed, not reproducible in-harness);
- LV2 has no buildable target in this environment, so its changes are compile-checked only.


jcelerier:
## Adversarial review pass: every `token_request` call site + buffer bounds

Walked every caller of the helpers this branch touches and checked that nothing can index outside a buffer. Findings, then what changed.

### The invariant that makes the audio side safe

For `x = offset·ratio/|speed|`:
- `physical_start = trunc(x) = floor(x)` (x ≥ 0 after this branch)
- `safe_physical_write_duration = floor(bufferSize − x)`
- `floor(x) + floor(bufferSize − x)` is `bufferSize − 1` when x is fractional and `bufferSize` when it is integral

so `physical_start + samples_to_write ≤ bufferSize` for **both** the sound nodes and `timings()`. And because `sample_info()` now takes `min(tick_dur, room)`, `samples_to_write` only ever *shrinks* — no consumer that sizes a temporary from it (the `alloca` in `sound_mmap::fetch_audio*`, `float_v`, faust's `alloca`) can overflow as a result of this branch.

`ot_offset = tk.offset + tick_ms − ot` with `ot ≥ 0` gives `ot_offset ≤ tk.offset + tick_ms ≤ buffer duration`, so `physical_start < bufferSize` whenever a token is actually emitted. The `ot == 0` case lands exactly at `bufferSize` but emits no token (`displacement == 0` → `tick == 0` → no tick), so `timings()` never hits its `start_sample >= bufferSize` error path.

### Latent crashes this branch removes

- **Pd**: `uint64_t(t.physical_start(...))`. With the old signed division that was `uint64_t(-256)` ≈ 1.8e19 on any backward tick, straight into `channel.reserve(that + req_samples)` → `bad_alloc`. Now non-negative.
- **MIDI timestamps**: `midi.hpp` derives `tick_start` from `physical_start` and clamps its stamps to `[tick_start, tick_end−1]`. With the old negative `physical_start` those went negative, and negative stamps reach VST3's `int32 sampleOffset` and CLAP's `uint32 header.time` (→ ~4e9). `tick_end` is bounded: `floor(x) + ceil(bufferSize − x) = bufferSize`, so stamps stay `≤ bufferSize − 1`.

### Bugs the review found — now fixed

**MIDI sample offsets were never rebased onto the tick's span.** A plug-in is handed `[tick_start, tick_start+samples)` and indexes from 0, but port timestamps are buffer-relative. This is wrong on the first buffer of *every* interval that doesn't start buffer-aligned, in both directions — and this branch makes VST3 honour the offset for audio, which would have left its MIDI path inconsistent with its audio path.

- **VST3**: `e.sampleOffset = mess.timestamp` → rebased and clamped to `[0, samples)`; readback undoes it. Threaded `tick_start`/`samples` into `dispatchMidi`/`readbackMidi`.
- **VST2**: already rebased but unclamped, so `deltaFrames` could go negative → clamped.
- **ysfx**: `ev.offset = mess.timestamp` → rebased and clamped; readback undoes it.
- **CLAP** (new commit, wasn't touched before): MIDI/note/MIDI2 event times passed through verbatim; parameter events clamped into the block but not rebased. All rebased + clamped, readback undone.

**A robustness hole I introduced.** Narrowing `musical_tick_duration <= 0` to `== 0` meant the walk direction was inferred from the musical duration alone. An interval with its own negative speed under an already-rewinding parent ticks *forward* while inheriting the parent's *decreasing* musical positions — the walk then ran with the clamps of the other direction and discarded every point. Now the tick's own direction decides, and a disagreement falls back to the single `prev_date` point. `metronome()` likewise requires a positive musical duration, not merely non-zero, before interpolating a position inside the tick. Both covered by `test_quantification_dates_direction_mismatch`, which fails without them.

### Verified safe, unchanged

- **avnd** (airwindows &c.) windows *and* rebases both control and MIDI ports on `[start, start+frames)` — safe by construction, which is why avnd never needed the fixes above.
- `read_audio_from_buffer{,_backward}` guard every file read with `pos >= 0 && pos < file_duration`; destinations are bounded by `samples_to_write`.
- `value_port` timestamps from `metronome.hpp` / `automation` / Patternist are all `≥ physical_start ≥ 0` and `< start+frames`.
- Patternist's step dates are `< tick_start + frames` (the walk breaks strictly before `date`), and its messages come out in increasing timestamp order, which VST3 and CLAP both require.

### Reported, not fixed

- **halp has its own copy of the quantification logic** (`halp/audio.hpp`), with the same `musical_tick_duration <= 0.` shortcut — so `VelToNote` / `RateLimiter` still fire one point per buffer on rewind. Frame offsets stay in `[0, dt)` so it is not a memory-safety issue, just wrong timing. It needs the same treatment as libossia's copy.
- **`halp::tick::pos_to_frame()`** divides by `end − start` with no guard: a paused tick makes that 0 → inf/NaN → UB on the `int` conversion. Currently no in-tree caller.
- **An interval whose own speed ≠ 1 crossing a boundary mid-buffer** gets an offset computed by mixing the parent and interval time frames, and can come out negative — the `token_request` constructor clamps it to 0, so it is a wrong position rather than a bad index. Pre-existing and symmetric between directions (the forward path has the identical expression); not touched.

Rebuilt clean and re-ran everything: libossia 39/40 (only the pre-existing `ScenarioAlgoTest`), Patternist 16/16.

PR #2165, opened 2026-07-30
fix(widgets): scale custom cursor hotspots on xcb for HiDPI

### LINKED ISSUE #2164: Cursor hotspot is off on HiDPI / UI Zoom > 100%, so hitboxes appear misplaced
## Symptom

With a HiDPI screen and/or **Settings > User interface > Graphical Zoom** above 100%, small
graphics items appear to have misplaced hitboxes: pointing at a widget activates whatever is
up and to the left of it, and you have to aim down-right of a control to actually hit it.

Easiest place to see it: hover a state in the scenario so its three creation crosses appear.
At 100% pointing at a cross gives the pointing-hand cursor as expected. At 200%, pointing
straight at a cross instead activates the state underneath (the cursor turns into the move
cursor), and the cross only responds when the pointer is visibly down and to the right of it.

## Cause

It is not the hitboxes — it is the **cursor image being drawn offset from the real pointer
position**, so aiming by the visible cursor misses.

`score::get_cursor` (`src/lib/score/widgets/SetIcons.cpp`) loads the `@2x` cursor asset once
`devicePixelRatio() >= 1.5` and tags the pixmap with `setDevicePixelRatio(2.)`. The hotspots
in `score::Skin` (`src/lib/score/model/Skin.cpp:141-172`) are in device-independent pixels,
which is what `QCursor` expects — Qt's own default hotspot is
`bitmap.width() / 2 / devicePixelRatio` (`qcursor.cpp`, `QCursorData::setBitmap`).

Qt's xcb backend, however, passes `QCursor::hotSpot()` straight to XRender together with the
raw device-pixel image, with no DPR conversion:

```cpp
// qtbase/src/plugins/platforms/xcb/qxcbcursor.cpp, QXcbCursor::createBitmapCursor
QPoint spot = cursor->hotSpot();
c = qt_xcb_createCursorXRender(m_screen, cursor->pixmap().toImage(), spot);
// -> xcb_render_create_cursor(conn, cursor, pic, spot.x(), spot.y()) on the 64x64 image
```

So a 64x64 `@2x` cursor gets its hotspot at half the intended offset, and the image is drawn
that much down-right of the true pointer.

## Measurements

Qt 6.4.2, X11, `Xft.dpi: 192` (2x desktop scale) plus Zoom 200% -> `devicePixelRatio() == 4`.
Read back from the X server with `XFixesGetCursorImage`:

| cursor | hotspot passed (logical) | image | X server hotspot | should be |
|---|---|---|---|---|
| `CursorPointer` | (12,10) | 64x64 | (12,10) | (24,20) |
| `CursorMove` | (15,15) | 64x64 | (15,15) | (30,30) |
| `CursorPointingHand` | (16,16) | 64x64 | (16,16) | (32,32) |

The gap is ~16 device px, and it grows with the effective devicePixelRatio, which is why it
is only noticeable above 100% zoom (at DPR 1 the base 32x32 asset is used and the hotspot is
correct).

This affects every custom cursor in score, not just the scenario, on any setup where
`devicePixelRatio() >= 1.5` — a HiDPI display alone is enough, Zoom just makes it worse.

## Scope

The xcb code above is unchanged from Qt 6.2 through current `dev` (6.11), so no Qt upgrade
fixes it and the workaround has to live in score. Other platforms honour the pixmap
devicePixelRatio for cursors and are left alone.

Related: #2048 (misplaced hitboxes on Windows) was closed as fixed; this one is the X11 case.

🤖 Investigated with [Claude Code](https://claude.com/claude-code)

edumeneses:
Closes #2164.

## Problem

On X11, custom cursors are drawn ~16 device pixels down and to the right of the real pointer
position whenever `devicePixelRatio() >= 1.5`, so aiming by the visible cursor misses: every
small control looks like its hitbox is misplaced. It is most obvious with **Graphical Zoom**
above 100%, because Zoom raises the effective devicePixelRatio.

`score::get_cursor` loads the `@2x` asset and tags it `setDevicePixelRatio(2.)`. The hotspots
in `score::Skin` are in device-independent pixels, which is what `QCursor` expects. Qt's xcb
backend then hands `QCursor::hotSpot()` to `xcb_render_create_cursor` together with the raw
device-pixel image, without converting it:

```cpp
// qtbase/src/plugins/platforms/xcb/qxcbcursor.cpp, QXcbCursor::createBitmapCursor
QPoint spot = cursor->hotSpot();
c = qt_xcb_createCursorXRender(m_screen, cursor->pixmap().toImage(), spot);
```

so on a 64x64 cursor the hotspot ends up at (16,16) instead of (32,32).

## Fix

Express the hotspot in the image's own pixels on xcb only. Other platforms honour the pixmap
devicePixelRatio for cursors, and the `devicePixelRatio() < 1.5` path is unaffected because
the pixmap DPR is then 1.

## Verification

Qt 6.4.2, X11, `Xft.dpi: 192` plus Zoom 200% -> `devicePixelRatio() == 4`. Hotspots read back
from the X server with `XFixesGetCursorImage`, in a running score:

| cursor | logical hotspot | image | before | after |
|---|---|---|---|---|
| `CursorPointer` | (12,10) | 64x64 | (12,10) | **(24,20)** |
| `CursorMove` | (15,15) | 64x64 | (15,15) | **(30,30)** |
| `CursorPointingHand` | (16,16) | 64x64 | (16,16) | **(32,32)** |

Before the fix, pointing at a state's creation cross activated the state underneath it; after,
hovering the cross's own pixels gives the pointing-hand cursor as expected. Checked that Zoom
100% is unchanged.

Note for anyone testing a second cursor in the same process: the xcb backend caches cursors on
`(shape, pixmap cacheKey, mask cacheKey)` and the hotspot is *not* part of that key, so
rebuilding a cursor from the same pixmap returns the cached one. Not an issue in score, where
each cursor comes from its own asset.

https://claude.ai/code/session_01TTUi76cnUWh7dGZC64LqbZ

### DIFF
--- a/src/lib/score/widgets/SetIcons.cpp
+++ b/src/lib/score/widgets/SetIcons.cpp
@@ -331,7 +331,11 @@ QCursor get_cursor(QString str, double hotspot_x, double hotspot_y)
   {
     img.load(str);
   }
-  cur = QCursor{img, (int)hotspot_x, (int)hotspot_y};
+  double hotspot_scale = 1.;
+  if(QGuiApplication::platformName() == QLatin1String("xcb"))
+    hotspot_scale = img.devicePixelRatio();
+
+  cur = QCursor{img, (int)(hotspot_x * hotspot_scale), (int)(hotspot_y * hotspot_scale)};
   return cur;
 }
 


src/lib/score/widgets/SetIcons.cpp
```diff
@@ -331,7 +331,18 @@ QCursor get_cursor(QString str, double hotspot_x, double hotspot_y)
   {
     img.load(str);
   }
-  cur = QCursor{img, (int)hotspot_x, (int)hotspot_y};
+  // The hotspots are given in device-independent pixels, which is what QCursor
```
jcelerier: can you ask claude to remove these slop comments :D 

also, can you try on a X11 session as well as a wayland one with XWayland (QT_QPA_PLATFORM=xcb to enforce xcb on wayland)? And also check if it doesn't need it with QT_QPA_PLATFORM=wayland.

thanks!

edumeneses: Sure. I'll add the directive to Claude's memory. I'll double-check on both platforms, but I think the comment above was wrong and I tested on Wayland before (but not X11). I'll see

edumeneses:
Comments removed — the diff is code only now, amended into the same commit.

Tested the three configurations you asked for. Qt 6.4.2 throughout, `Skin/Zoom` 200%.

**1. X11 session, xcb** — needed. `Xft.dpi: 192` (2x desktop scale) x Zoom 200% => `devicePixelRatio() == 4`. Hotspots read back from the X server with `XFixesGetCursorImage`, in a running score:

| cursor | before | after | image |
|---|---|---|---|
| `CursorPointer` | (12,10) | (24,20) | 64x64 |
| `CursorMove` | (15,15) | (30,30) | 64x64 |
| `CursorPointingHand` | (16,16) | (32,32) | 64x64 |

**2. Wayland session (nested mutter 46.2) with `QT_QPA_PLATFORM=xcb`, i.e. XWayland** — needed, and behaves exactly like a real X11 session. `platformName()` is `xcb` so the patch applies; `devicePixelRatio() == 2`, and the XWayland server reports the patched `CursorPointer` at (24,20). Control on the same display, building the cursor by hand from `cursor_pointing_hand@2x.png`:

```
hotspot passed = 16    X server sees: image 64x64  hotspot (16,16)  image centre (32,32)  <- OFF CENTRE
hotspot passed = 32    X server sees: image 64x64  hotspot (32,32)  image centre (32,32)  <- centred, correct
```

**3. Same session with `QT_QPA_PLATFORM=wayland`** — not needed, and the patch correctly stays out of the way. `platformName()` is `wayland`, `devicePixelRatio() == 2`, the `@2x` pixmaps are loaded (64x64, dpr 2), and the hotspots score hands to Qt are still the plain logical values, unchanged from before the patch:

```
platformName()           : wayland
  bitmap cursor on score::View:       pixmap 64x64 dpr 2, hotSpot (12,10)
  bitmap cursor on score::TimeSpinBox: pixmap 64x64 dpr 2, hotSpot (16,16)
```

That matches qtwayland: the cursor hotspot there is in *surface* coordinates, not buffer pixels. `QWaylandInputDevice::setCursor` takes `cursor->hotSpot()` as-is and only touches it when the dpr has a fractional part, and the themed-cursor path divides by `bufferScale`:

```cpp
mCursor.hotspot = cursor ? cursor->hotSpot() : QPoint();
mCursor.bitmapScale = int(dpr);
if (mCursor.bitmapScale < dpr)          // fractional only
    mCursor.hotspot *= dpr / mCursor.bitmapScale;
```

So scaling on Wayland would double the hotspot — hence the `platformName() == "xcb"` gate rather than an unconditional multiply.

One caveat on what was *not* verified: on the Wayland leg I checked that the patch is a no-op (same values as before) rather than visually confirming cursor alignment, since there is no equivalent of `XFixesGetCursorImage` to ask the compositor where the hotspot landed.

Also worth knowing if anyone tests cursors interactively: the xcb backend caches cursors on `(shape, pixmap cacheKey, mask cacheKey)` and the hotspot is *not* part of that key, so rebuilding a cursor from the same pixmap in one process hands back the cached one. Not an issue in score, where each cursor has its own asset, but it silently invalidated my first test run.

— posted by Edu's Claude Code assistant

edumeneses:
Humanly tested on 2 machines, one running X11 and another wayland.

jcelerier:
Hmm it's strange that you need to put zoom at 200% if you already have Xft.dpi: 192. thanks for testing, i'll merge when build is green !

edumeneses:
Ah, I normally use Zoom at 100%. I changed to 200% for a random reason during unrelated SAT tests (I don't even remember why; I think the person I wanted to show something to couldn't see properly).

jcelerier:
@claude this looks like something that should eventually be fixed upstream in Qt, prepare a short patch

Issue #919, opened 2019-04-04
add a shortcut to remove trigger


avilleret:
currently I have to right clic and go to events menu to disable the trigger
I used to use suppr when selecting trigger on the scenario (but that was buggy, when the trigger was selected from the inspector, suppr key didn't work)

bltzr:
suppr seems the way to go - It used to work this way, right (when you redesigned it, 2 years ago ±), right ?

jcelerier:
fixed in 2.2.1

PR #2171, opened 2026-08-01
yocto: add meta-ossia layer for building flashable score images

jcelerier:
Adds a Yocto/OpenEmbedded layer under `cmake/Deployment/Linux/Yocto/` that builds ossia score into a bare-bones flashable image: kernel, init, audio and GPU stacks, and score with every plugin and addon that cross-compiles. No desktop environment — the target is an appliance rather than a small image.

> **Stacked on top of #2170.** The distro drops `x11` from `DISTRO_FEATURES`, so score is built with no X11 or xcb headers at all, which needs the X11-optional changes in that PR. Review #2170 first; this PR's own diff is only the new layer.

## Verified

Built and booted end to end on qemux86-64 — 6369 tasks, 901 packages. The OS reaches the Multi-User target in **7.1 s**, and score starts, renders its full UI and is interactive: with `/dev/dri` present, `ossia-score-launch` selects eglfs on DRM/KMS and Qt brings up RHI on LLVMpipe with no X11 or xcb anywhere in the image.

Artefacts: 186 MB `wic.xz`, 230 MB `ext4.zst`; the stripped binary is 83 MB, since nearly everything is linked into it.

### score is not yet stable in this configuration

Running under runqemu's default **256 MB** (214 MB usable), score does not survive:

```
[  6.40] systemd[1]: Started ossia score.
[ 43.02] ossia-score.service: Main process exited, code=dumped, status=11/SEGV
[ 45.23] ossia-score.service: Scheduled restart job, restart counter is at 1.
[112.01] ossia-score.service: Main process exited, code=killed, status=9/KILL
[112.01] ossia-score.service: Failed with result 'oom-kill'.
```

It segfaults once, and the restarted instance — the one that produced the screenshot below — is later OOM-killed. The OOM is plausibly just the memory ceiling (score's binary alone is 83 MB, before Qt and LLVMpipe compositing 1280×800); the **SEGV is unexplained** and is a separate, earlier failure. Neither has been root-caused yet, and neither is specific to anything in this layer — but they are open questions, not a clean bill of health.

Not verified: real hardware, the Raspberry Pi config, any GPU-accelerated path (qemu has no GPU, hence LLVMpipe), and long-run stability.

## What is pinned, and why

| | |
|---|---|
| openembedded-core / bitbake | `yocto-5.3.4` (**whinlatter**) |
| meta-openembedded | `whinlatter` |
| meta-qt6 | `6.12` → Qt **6.12.0** |
| rest | Boost **1.89.0**, GCC **15**, CMake 4.1.2, ffmpeg 8.0, glibc |

whinlatter is not arbitrary: boost 1.89 falls inside the window that `3rdparty/libossia/cmake/deps/boost.cmake` probes with `find_package(Boost 1.N EXACT)`. Outside it, that code downloads a boost tarball at configure time — fatal under BitBake's offline `do_configure`. wrynose (1.90) also works; oe-core master ships 1.91 and would need ossia/libossia#918.

There is no `poky/whinlatter` branch — poky stops at walnascar — so the stack is openembedded-core plus bitbake directly, and the distro config lives in the layer rather than coming from meta-poky.

## Source

The default recipe builds the score checkout this layer ships inside, via `externalsrc`, because development targets master. `ossia-score_3.8.2.bb` remains for reproducible builds behind `PREFERRED_VERSION`. Worth knowing: a plain `gitsm://` fetch of ossia/score would silently build with **no addons** — those are ~20 separate repositories that `ci/common.deps.sh` clones.

## Feature coverage

Everything that cross-compiles is built. Two addons are excluded on technical grounds rather than size: `score-addon-onnx` `FetchContent`s a *prebuilt x86_64* onnxruntime during configure, and `score-addon-ultraleap` pulls a proprietary SDK — neither works offline nor cross-compiles.

`lv2`, `faust` and `ysfx` self-disable for want of OE recipes (they use `find_package` + `if(NOT TARGET) return()`), so the build succeeds without them. The README lists what each would need.

## Relationship to the other PRs

- **#2170** — required; this is stacked on it.
- **ossia/libossia#919** — `find_package(rapidfuzz CONFIG REQUIRED)` aborts before its own vendored fallback can run, so this layer carries a `rapidfuzz-cpp` recipe. That recipe becomes optional once #919 lands.
- **#2169** — makes the `do_install` cleanup of xtensor's stray Jupyter/xeus files redundant. Harmless either way; it targets paths that will simply stop existing.

Configuration follows the existing distro-packaging builds (`ci/debian.trixie-system.build.sh`, `ci/nix.build.nix`, the Flatpak manifest): system libraries, FHS layout, static plugins, `-O3`, plus realtime scheduling limits for the audio engine.

### DIFF
--- a/cmake/Deployment/Linux/Yocto/README.md
+++ b/cmake/Deployment/Linux/Yocto/README.md
@@ -58,8 +58,17 @@ stall.
 ### Running it
 
 ```bash
-kas shell kas/whinlatter-qemux86-64.yml -c "runqemu nographic ossia-score-image"
+kas shell kas/whinlatter-qemux86-64.yml -c \
+  "runqemu kvm nographic slirp snapshot ossia-score-image qemuparams='-m 2048'"
 ```
+
+Two arguments there are not optional in practice:
+
+* **`kvm`** — runqemu does *not* enable KVM unless you ask. Without it every guest
+  instruction is interpreted and the whole system is roughly 3x slower, which
+  distorts any measurement you take (it inflated our first power-to-GUI figure
+  from 6.1s to 18.4s, most of it in software rasterisation).
+* **`-m 2048`** — the 256 MB default is not enough; see the sizing note below.
 
 then on the target:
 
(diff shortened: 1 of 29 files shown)


jcelerier:
Moved to https://github.com/ossia/meta-ossia.

A BSP layer belongs in its own repo: it can be added to `bblayers.conf` without cloning score, and it versions independently of score releases. The layer builds the release tarball by default, or a local checkout via `SCORE_SRC_ROOT`.

The upstream fixes this work produced are already merged separately: #2169, #2170, ossia/libossia#918 and ossia/libossia#919.

PR #2181, opened 2026-08-08
media: link the codec libraries ossia SDK 38's ffmpeg needs

jcelerier:
Companion to **ossia/sdk#30** (ffmpeg 9 + network/TLS + hardware decoding + codecs). Without this, a static score against SDK 37 does not link.

### Why

SDK 37's `libavcodec`/`libavformat` reference dav1d, x264, x265, opus, vpx, webp (+`sharpyuv`, `webpmux`), SVT-JPEG-XS, mp3lame, libxml2, SRT (which links the SDK's openssl) and freetype+harfbuzz for `drawtext`.

`cmake/modules/FindFFmpeg.cmake` imports only the `av*` archives — it runs pkg-config for the include/library *directories* and then `find_library()`s each component, so `Libs.private` never reaches the link line. Nothing else in the tree names these libraries, so the link fails with undefined references.

### Why it can't break the matrix

- The whole block is inside the existing `if(OSSIA_SDK)` — distro and Homebrew ffmpeg builds are untouched.
- Every library is *looked up*, not required. SDK 36 has none of them → nothing is added. Even inside SDK 37 some are target-dependent (Windows-on-ARM has no OpenSSL mingw target, so libsrt is built without encryption there and no libssl exists).
- `NO_DEFAULT_PATH` on every lookup, searching only the SDK's own directories, so a host `libssl`/`libfreetype` can never be picked up silently.
- The WASM path is a separate branch (`if(EMSCRIPTEN)`) and is not touched.

### Ordering

`PUBLIC`, because score-plugin-gfx links `avcodec`/`avformat` directly and lists `score_plugin_media` after them, so it inherits these in the order a static link needs. Within the list dependants precede dependencies (webp → sharpyuv, srt → ssl/crypto), and `Threads::Threads` is last because raw archive paths carry no dependency info for CMake to order around.

### Validation

| what | result |
|---|---|
| Windows, real SDK 37 prefix on a mingw box: program calling `avdevice_register_all` + `avformat_network_init` and iterating every codec, filter, demuxer, muxer and protocol | links with exactly this list, runs clean (exit 0) |
| its import table | system DLLs only — no `libcuda`, `vulkan-1` or `amfrt64`, so the SDK's hardware acceleration is confirmed dlopen-only |
| CMake logic vs. SDK 37 Linux/Windows layout | all 16 resolved, in order |
| vs. SDK 36 layout | nothing added |
| vs. macOS layout (`$OSSIA_SDK/lib`) | all resolved |
| vs. an SDK with no openssl | system `libssl` correctly **not** picked up |

The Windows system import libraries added under `if(WIN32)` (mfuuid/ole32/strmiids for MediaFoundation, secur32/ncrypt/crypt32 for schannel, ws2_32/wsock32 for SRT, bcrypt, psapi/uuid/oleaut32/shlwapi/gdi32 for libavdevice) were read off the `Libs:` lines of the `.pc` files the SDK installs.

### DIFF
--- a/ci/appimage.deps.sh
+++ b/ci/appimage.deps.sh
@@ -19,7 +19,7 @@ else
   export CPU_ARCH_SUFFIX="-x86_64"
 fi
 
-wget -nv "https://github.com/ossia/sdk/releases/download/sdk37/sdk-linux${CPU_ARCH_SUFFIX}.tar.xz"
+wget -nv "https://github.com/ossia/sdk/releases/download/sdk38/sdk-linux${CPU_ARCH_SUFFIX}.tar.xz"
 tar xaf sdk-linux${CPU_ARCH_SUFFIX}.tar.xz
 rm -rf  sdk-linux${CPU_ARCH_SUFFIX}.tar.xz
 
--- a/ci/osx.package.deps.sh
+++ b/ci/osx.package.deps.sh
@@ -33,7 +33,7 @@ brew install gnu-tar ninja xz
 wget -nv "https://github.com/jcelerier/cninja/releases/download/v3.7.9/cninja-v3.7.9-macOS-$MACOS_ARCH.tar.gz" -O cninja.tgz &
 
 SDK_ARCHIVE=sdk-macOS-$MACOS_ARCH.tar.xz
-wget -nv https://github.com/ossia/sdk/releases/download/sdk37/$SDK_ARCHIVE -O "$SDK_ARCHIVE"
+wget -nv https://github.com/ossia/sdk/releases/download/sdk38/$SDK_ARCHIVE -O "$SDK_ARCHIVE"
 
 sudo mkdir -p "/opt/ossia-sdk-$MACOS_ARCH/"
 sudo chown -R "$(whoami)" /opt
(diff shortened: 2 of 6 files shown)


jcelerier:
Rebased onto **SDK 38** (ffmpeg 9), which is what these codec-linking changes were written against — SDK 37 ships ffmpeg 8.1, whose libavcodec/libavformat do not pull in the extra libraries added here.

Updated: `ci/appimage.deps.sh`, `ci/osx.package.deps.sh`, `ci/wasm.deps.sh`, `ci/win32.deps.sh`.

Two references are deliberately unchanged:

- `ci/win32.msvc.deps.sh` stays on `sdk35` — the MSVC archive is no longer built (sdk37 publishes only linux/macOS/mingw/wasm), so sdk35 is the last release containing `sdk-msvc-x86_64.7z`.
- `tools/fetch-sdk.sh` stays on `sdk37` — it downloads from `ossia/score-sdk`, a *different* repository that mirrors releases separately and whose newest tag is `sdk37`. Bumping it would 404 and break local setup via `tools/developer.sh`. It needs an `sdk38` release published there first.

Note the `sdk38` tag currently has **no release assets** — the tag push triggered the SDK build and it is still running, so CI here will fail to download until those artifacts are published.

PR #2187, opened 2026-08-10
avnd: register Midi Humanize, and pin the Midi Envelope fixes with tests

jcelerier:
**Draft: stacked on #2182, and blocked on celtera/avendish#188.**

Rebased onto `feature/win32-fixes-08`, which already registers `Midi Envelope`, carries the execution index fix (`avnd: fix an index offset issue for execution`) and has the base envelope test. This PR is now pure additions on top of that — no duplication.

Still blocked on the submodule: celtera/avendish#188 adds `MidiHumanizer.hpp` **and** fixes a set of correctness bugs in `Midi Envelope`. The tests here pin those fixes, so they only pass once `3rdparty/avendish` is bumped past it. I deliberately did not pin the submodule to my unmerged branch head — a pin to a commit that may be squashed away is worse than an obvious missing bump.

## Contents

- Registers `midihumanizer`.
- The `Midi Humanize` test suite.
- Extends the existing envelope test suite with a regression test for each fix in avendish#188.

## Tests

**Midi Humanize** — 14 cases / ~171k assertions. The centrepiece is a direct **DFT check of the noise exponent**: the object's own timing deviations are collected, then their periodogram is regressed against `log(2·sin(ω/2))` — the exact spectrum of the fractional-difference filter, so no low-frequency approximation — and must show the requested `1/f^β` slope. Measured **-0.010 / -1.012 / -1.999** for β = 0 / 1 / 2, held to ±0.10. That is the object's entire reason to exist (Hennig et al. 2011: human rhythmic fluctuation is long-range correlated, and listeners prefer it over the white noise every other humanizer produces), so it is worth pinning numerically rather than by eye.

Also covered: unit variance across colours, so the amount control means the same thing at any β; note-off/note-on pairing and note-length preservation; chord grouping and strum; velocity clamping; seed reproducibility; and a "never hangs a note" suite covering seed change mid-note, all-notes-off with a full queue, and more overlapping notes than the tracking table holds.

**Midi Envelope** — extended to 17 cases / 129 assertions, adding regression tests for the stranded-voice bugs (both the note-off path and the `advance()` path), the mono-priority spurious retrigger, voice-count shrink, the held-note cap, legato semantics and the smoothing time step.

I verified these fail against the unfixed object rather than assuming: reverting the header shows every one of them failing with the expected wrong values.

Two deliberate test choices: the saturation case asserts `held <= 0` per pitch (never *left sounding*) rather than perfectly balanced, since past the queue and tracking limits notes may legitimately be dropped and an unmatched note-off is inert; and the legato test discriminates on whether the voice is still at the sustain level, because a non-`Reset` restart resumes from the current level and so cannot be detected from the level alone.

https://claude.ai/code/session_01S5kNpDQVQfbvHQBtv8LCNv

### DIFF
--- a/src/plugins/score-plugin-avnd/CMakeLists.txt
+++ b/src/plugins/score-plugin-avnd/CMakeLists.txt
@@ -158,6 +158,7 @@ add_library(
   score_plugin_avnd
 
   "${AVND_FOLDER}/examples/Advanced/Utilities/ADSR.hpp"
+  "${AVND_FOLDER}/examples/Advanced/MidiScaler/MidiEnvelope.hpp"
   "${AVND_FOLDER}/examples/Advanced/Audio/AudioFilters.hpp"
   "${AVND_FOLDER}/examples/Advanced/Audio/Bitcrush.hpp"
   "${AVND_FOLDER}/examples/Advanced/Audio/Convolver.hpp"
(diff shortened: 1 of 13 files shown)


jcelerier:
Folded into #2182 — the work belongs on a single branch. #2182 is now rebased on current master (497d8007) with the avendish submodule bumped past celtera/avendish#188, which is merged.

PR #2195, opened 2026-08-10
Fix FFmpeg video decoder packet and frame draining

darioriccio:
## Summary

- retry the same AVPacket when `avcodec_send_packet` reports decoder back-pressure
- drain every frame made available by a packet before reading the next packet
- preserve output ordering when one packet yields multiple frames and use the same path for EOF flushing

## Root cause

The previous loop consumed the next demuxed packet after `avcodec_send_packet` returned `EAGAIN`, even though FFmpeg requires the caller to receive pending frames and retry the unchanged packet. Inter-frame HEVC streams then lost reference packets and repeatedly emitted `Could not find ref with POC`, eventually freezing.

## Verification

- standalone send/receive probe on the affected HEVC stream: 400/400 frames decoded with the corrected loop, versus 252/400 and reference-frame errors with the previous loop
- integrated macOS Domeport Pro build using the background `VideoDecoder` path and hardware decoding disabled: an 8K HEVC video advanced continuously, produced distinct rendered frames, and emitted no HEVC POC/reference errors
- EOF drain uses the same multi-frame receive path

This is intentionally limited to the background decoder send/receive contract; it does not change renderer selection or UI behavior.

### DIFF
--- a/src/plugins/score-plugin-media/Video/VideoDecoder.cpp
+++ b/src/plugins/score-plugin-media/Video/VideoDecoder.cpp
@@ -217,31 +217,79 @@ LibAVDecoder::open_hwdec(const AVCodec& detected_codec) noexcept
 
 ReadFrame LibAVDecoder::enqueue_frame(const AVPacket* pkt) noexcept
 {
-  auto frame = m_frames.newFrame();
+  auto receive = [this]() -> ReadFrame
+  {
+    auto frame = m_frames.newFrame();
+    auto read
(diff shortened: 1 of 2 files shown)


jcelerier:
hello ! can you check if this works fine on top of https://github.com/ossia/score/pull/2121 ?  as it's a pretty big refactor of all the video part to enable better performance

jcelerier:
@darioriccio do you have a file that produced the error ?

jcelerier:
thanks ! rebased and merged in master

PR #2196, opened 2026-08-10
Allow custom apps to select video decoding via environment

darioriccio:
## Summary

- read an optional `SCORE_VIDEO_DECODING_METHOD` environment variable at startup
- match the requested decoder case-insensitively against the decoders available in the build
- preserve the existing persisted setting when the variable is absent or invalid

This lets downstream applications such as Domeport Pro request `Auto` without changing score defaults or adding application-specific code to score.

## Verification

The combined Domeport Pro macOS artifact built successfully in run 31439148108. With the persisted hardware-decoder preference deliberately set to `None`, the packaged process had `SCORE_VIDEO_DECODING_METHOD=Auto` and continuously played the affected HEVC 4320x2160 file at about 52-54 FPS. This confirms the application environment overrides the stored preference while leaving the global default unchanged.

### DIFF
--- a/src/plugins/score-plugin-gfx/Gfx/Settings/Model.cpp
+++ b/src/plugins/score-plugin-gfx/Gfx/Settings/Model.cpp
@@ -159,6 +159,22 @@ Model::Model(
 {
   score::setupDefaultSettings(set, Parameters::list(), *this);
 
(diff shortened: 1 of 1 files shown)


jcelerier:
thanks !

PR #2244, opened 2026-08-30
gfx: six defects found while building GPU/video test coverage

jcelerier:
Six defects found while building out GPU/video test coverage, extracted from the
feature stack so they can land on master independently of it.

| commit | defect |
|---|---|
| `42eb64fc71` | compute, geometry-filter and VSA shaders could not be dropped into a score — three drop handlers were implemented but never registered, and two of them declared four overrides while defining only two, leaving an undefined vtable |
| `b263758a80` | a malformed `.vox` aborted the process instead of failing the load |
| `473e3840fc` | MIDI bytes were dropped, audio outlets were never cleared, and four scripting-API calls aborted on wrong-typed input |
| `f018e4908f` | creating a device with incomplete settings aborted the application — rapidjson accessors are `RAPIDJSON_ASSERT`, so absent or wrong-typed settings abort rather than throw |
| `f0065814a9` | video colour was decoded with a full-range chroma centre of 0.5 instead of 128/255, and high-bit-depth sources were normalised by `2^n - 1` instead of swscale's `255 * 2^(n-8)`; the image cache also evicted on first release |
| `3426134607` | devices dereferenced the execution context from their destructors |

## On the colour fix

The decoder suite's tolerance goes from `kTol = 3` to **`kTol = 0`** — exact
agreement with the reference — and assertion count rises 382 → 686 across both
backends. Negative-controlled twice: forcing the scale back to 1.0 produces 152
failures, reverting the whole `decoders/` directory produces 238.

Two corrections were made to the fix along the way, both recorded in the commit:
8→10-bit is a *shift* per SMPTE 274M (`235 → 940`), not bit replication; and one
test expectation was itself wrong (`cr-hi` wanted G = 68, which comes from a 0.5
centre — ffmpeg and exact BT.709 arithmetic both give 69).

Deliberately **not** included: the encoder-side mirror of the chroma-offset
defect. The correct encode offset is depth-dependent (`2^(n-1)/(2^n - 1)`, so
0.501961 at 8 bits and 0.500489 at 10), which one mat4 cannot serve. It needs its
own change and its own test.

## Verification

Built and run at `RelWithDebInfo`: **97%, 4 failures of 119**, all four accounted
for and none related to these changes:

- `test_unit_clap_scan` — the pre-existing `TimeVal` QVariant metatype failure
- `test_library_publish` — UBSan inside vendored `3rdparty/libpd/pure-data/`; Pure Data's dispatch idiom, wants a scoped suppression rather than a patch
- `test_integration_media_trim`, `test_integration_unused_files` — artifacts of this build tree living under `/tmp`: both are sandboxed with bwrap, which mounts `--tmpfs /tmp` and hides the binaries (`bwrap: execvp ...: No such file or directory`)


src/lib/score/tools/Timers.cpp
```diff
@@ -89,10 +89,18 @@ class HighResolutionTimerPrivate
   int timerId{-1};
   bool accurate{};
 
+  // A frequency of zero -- reachable from a device's Rate setting -- makes both
+  // derived intervals inf, and converting inf to an integer is undefined; it
+  // arrived at startTimer as a negative interval.
+  static double sanitizeFrequency(double freq) noexcept
+  {
+    return (freq > 0. && std::isfinite(freq)) ? freq : 1.;
```
jcelerier: ossia::safe_isfinite

src/plugins/score-plugin-gfx/Gfx/GPhoto2/GPhoto2Device.cpp
```diff
@@ -1579,6 +1580,6 @@ void JSONReader::read(const Gfx::GPhoto2::GPhoto2Settings& n)
 template <>
 void JSONWriter::write(Gfx::GPhoto2::GPhoto2Settings& n)
 {
-  n.model = obj["Model"].toString();
-  n.port = obj["Port"].toString();
+  Gfx::readJsonField(obj, "Model", n.model);
```
jcelerier: This should be a general utility in score's serialization code, not specific to gfx::. For instance: "score::parseJsonField" to indicate we're going to try parsing instead of forcing through rapidjson.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/ColorSpace.hpp
```diff
@@ -54,7 +58,7 @@ namespace score::gfx
     1.000000000000000,  1.000000000000000,  1.000000000000000, 0.0,\n\
     0.000000000000000, -0.344136286201022,  1.772000000000000, 0.0,\n\
     1.402000000000000, -0.714136286201022,  0.000000000000000, 0.0,\n\
-   -0.701000000000000,  0.529136286201022, -0.886000000000000, 1.0)\n"
+   -0.703749019607843,  0.531211330460634, -0.889474509803922, 1.0)\n"
```
jcelerier: make sure that we have tests that validate that all of this is good with actual test footage, for instance generated through ffmpeg.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/ColorSpace.hpp
```diff
@@ -134,7 +138,7 @@ namespace score::gfx
     1.0,  1.0,  1.0, 0.0,\n\
    -1.0,  1.0, -1.0, 0.0,\n\
     1.0,  0.0, -1.0, 0.0,\n\
-    0.0, -0.5,  0.5, 1.0)\n"
+    0.000000000000000, -0.501960784313725,  1.003921568627451, 1.0)\n"
```
jcelerier: the last one looks like a big jump, triple-check the matrix. look for the original source I used to understand the previous matrix. validate everything with actual tests.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/RGBA.hpp
```diff
@@ -30,17 +31,23 @@ struct PackedDecoder : GPUVideoDecoder
...
       , decoder{d}
       , filter{std::move(f)}
       , invertY{invertY}
+      , nearestSampling{nearest}
   {
   }
   QRhiTexture::Format format;
   int bytes_per_pixel{}; // bpp/8 !
   bool invertY{}; // Mainly useful for Spout texture import
+  /// Formats whose texel is reassembled from several bytes (big-endian 16-bit,
```
jcelerier: don't add multi-line comments when adding a single field. its usage from the commit is enough

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/RGBA.hpp
```diff
@@ -332,8 +340,13 @@ struct RGB24Decoder : GPUVideoDecoder
     const auto w = decoder.width, h = decoder.height;
 
     {
-      // Create a texture
-      auto tex = rhi.newTexture(QRhiTexture::R8, QSize{w * 3, h}, 1, QRhiTexture::sRGB);
+      // Create a texture. NOTE: no QRhiTexture::sRGB — this is a *data* texture
```
jcelerier: remove these slop comments

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/RGBA.hpp
```diff
@@ -442,5 +456,86 @@ struct RGB48Decoder : GPUVideoDecoder
   }
 };
 
+// RGBA / BGRA 16-bit-per-channel (rgba64le / bgra64le). The data is 16-bit
```
jcelerier: simplify the comments, don't ever mention how the issue was found or how the previous code worked in comments.

src/plugins/score-plugin-gfx/Gfx/GfxExecContext.hpp
```diff
@@ -25,6 +28,13 @@ class SCORE_PLUGIN_GFX_EXPORT GfxExecutionAction final
   void setEdge(port_index source, port_index sink, Process::CableType t);
   void endTick(const ossia::audio_tick_state& st) override;
 
+  // Devices hold a raw GfxExecutionAction* and reach ui through it from their
```
jcelerier: remove this comment

src/plugins/score-plugin-gfx/Gfx/GfxInputDevice.hpp
```diff
@@ -39,6 +39,11 @@ class simple_texture_input_parameter : public ossia::gfx::texture_parameter
 {
   GfxExecutionAction* context{};
 
+protected:
+  std::shared_ptr<bool> context_alive;
+
+public:
```
jcelerier: duplicate public:

src/plugins/score-plugin-js/JS/Qml/EditContext.device.cpp
```diff
@@ -243,6 +243,16 @@ void EditJsContext::createDevice(QString name, QString uuid, QJSValue obj)
   Device::DeviceSettings set;
   set.name = name;
   set.protocol = UuidKey<Device::ProtocolFactory>::fromString(uuid);
+  // The protocol has to exist whatever shape the settings argument has: this
```
jcelerier: remove this comment

src/plugins/score-plugin-js/JS/Qml/EditContext.device.cpp
```diff
@@ -371,6 +374,16 @@ void EditJsContext::createAddress(QString addr, QString type)
     return;
 
   auto& plug = doc->plugin<Explorer::DeviceDocumentPlugin>();
+
+  // Adding an address under a device that does not exist reaches
```
jcelerier: remove this comment

src/plugins/score-plugin-js/JS/Qml/EditContext.introspection.cpp
```diff
@@ -39,7 +39,13 @@ QVariant EditJsContext::availableProcessesAndPresets() const noexcept
   v.reserve(20000);
   auto& ctx = score::GUIAppContext();
   auto& procs = ctx.interfaces<Process::ProcessFactoryList>();
-  auto& lib = ctx.panel<Library::ProcessPanel>().processWidget().processModel();
+  // panel<T>() is SCORE_ABORT when the panel is absent, and it is absent under
```
jcelerier: remove this comment

src/plugins/score-plugin-js/JS/Qml/QmlObjects.cpp
```diff
@@ -279,7 +281,13 @@ void AudioOutlet::setChannel(int i, const QJSValue& v)
   if(i + 1 > std::ssize(m_audio))
     m_audio.resize(i + 1);
 
-  int n = v.property("length").toNumber();
+  // A non-array has no length: toNumber() gives NaN, and converting that to int
```
jcelerier: remove this comment

src/plugins/score-plugin-js/JS/Qml/QmlObjects.hpp
```diff
@@ -741,12 +744,17 @@ class SCORE_PLUGIN_JS_EXPORT MidiInlet : public Inlet
     m_midi.clear();
     for(const libremidi::ump& mess : arr)
     {
-      auto m1 = libremidi::midi1_from_ump(mess);
-      const auto N = mess.size();
-      QVector<int> m;
-      m.resize(N);
+      // ump::size() is the packet's WORD count, m1.bytes is the MIDI-1 byte
```
jcelerier: remove this comment

src/lib/score/serialization/JSONParse.hpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+  return false;
+}
+
+//! Narrowing a double that does not fit the destination is undefined, so the
+//! range is part of the type check.
+template <typename T>
+bool parseJsonNumber(double d, T& out) noexcept
+{
+  if(!std::isfinite(d))
```
jcelerier: ossia::safe_isfinite

src/lib/score/tools/Timers.cpp
jcelerier: Done — `ossia::safe_isfinite` + `<ossia/math/safe_math.hpp>`.

Also found while there: `sanitizeFrequency` only bounded the low end. `GfxContext.cpp:324` is not `qBound`ed like :310, so a Rate of 1e12 gives `intervalNs == 0` and spins the timer thread. Now clamped both ways.

src/lib/score/serialization/JSONParse.hpp
jcelerier: Done here too — `ossia::safe_isfinite`, same include.

Note the surrounding range check is doing separate work: `IsNumber()` is a type check, not a range check, so `static_cast<long long>(GetDouble())` on an out-of-range double is still UB. That was the exact UB this PR removes elsewhere and the first version of this helper reintroduced it.

src/plugins/score-plugin-gfx/Gfx/GPhoto2/GPhoto2Device.cpp
jcelerier: Done. `Gfx/SettingsJson.hpp` is deleted; the helpers now live in `src/lib/score/serialization/JSONParse.hpp` as `score::parseJsonField`, all 9 call sites migrated, header registered in `src/lib/CMakeLists.txt`.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/ColorSpace.hpp
jcelerier: Done — new `tests/unit/VideoColorSpaceTest.cpp`, 13221 assertions, two halves:

1. every matrix re-derived from first principles, hermetic and exact;
2. **real ffmpeg footage** — clips generated at run time, the shader's own macros replayed on the bytes ffmpeg wrote, compared against ffmpeg's own decode of those same bytes.

Covers `yuv444p{,10,12,16}le` x BT.601/709/2020 x tv/pc, `vuya`, `xv30le`, `p410le`, `gbr{,a}p{10,12,16}le`, `rgb48le`, `bgr48le`, `rgba64le`, `bgra64le`, plus the MSB-alignment identity and the Y210 texel mapping. Negative-controlled: pre-fix constants turn 5 of 8 cases red.

One honest gap: the subsampled MSB-aligned formats are validated by exact layout identity rather than through swscale, because **swscale's own 10-bit subsampled input path is imprecise** — it renders limited-range white as 253 where 8-bit 4:2:0, 10-bit 4:4:4 and zimg all give 255. Pinning that would pin swscale's error.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/ColorSpace.hpp
jcelerier: Your instinct was right, and it is worse than a big jump — **that matrix carries two fixes and the commit only explained one.**

The other full-range matrices move by 1/510 (the 0.5 -> 128/255 recentring). YCgCo full moves its blue offset by **0.5**. `B = Y - Cg - Co` carries the centring term *twice* where `G = Y + Cg` carries it once, so B's offset must be `2c` — it shipped as `c`. That is a factor-2 error in blue, independent of the recentring.

The limited-range YCgCo matrix pins the intended convention: its offsets are `-16/219`, `-16/219 - 128/224`, `-16/219 + 2*(128/224)` — the doubling is there and always has been. So `1.003921568627451 = 2*128/255` is correct.

On the original source: the header cites softpixel and vlc-qt's `GlPainter.cpp:48`. Both are BT.601-only and cover neither YCgCo nor full range — consistent with that row having been extrapolated by hand and losing the factor of 2.

All 13 matrices re-derived: 208 coefficients match to <1e-14. Commit message now states both fixes.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/RGBA.hpp
jcelerier: Done — cut to the single line of fact.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/RGBA.hpp
jcelerier: Done.

src/plugins/score-plugin-gfx/Gfx/Graph/decoders/RGBA.hpp
jcelerier: Done. Applied the rule PR-wide rather than only here — also stripped from `ColorSpace.hpp` (two blocks), `P010.hpp`, `Y210.hpp`, `EditContext.ui.cpp` and `CPUNode.cpp`. Rationale moved into the commit messages, which were rewritten.

src/plugins/score-plugin-gfx/Gfx/GfxExecContext.hpp
jcelerier: Done — cut to a two-line `//!`.

src/plugins/score-plugin-gfx/Gfx/GfxInputDevice.hpp
jcelerier: Done — removed, along with a stray `protected:` on the same class.

src/plugins/score-plugin-js/JS/Qml/EditContext.device.cpp
jcelerier: Done.

src/plugins/score-plugin-js/JS/Qml/EditContext.device.cpp
jcelerier: Done.

src/plugins/score-plugin-js/JS/Qml/EditContext.introspection.cpp
jcelerier: Done.

src/plugins/score-plugin-js/JS/Qml/QmlObjects.cpp
jcelerier: Done. Separately, `setChannel` bounded `n` but not `i`, so `setChannel(2000000000, [])` resized a `QVector<QVector<double>>`; and `n <= INT_MAX` is not much of a bound at 17 GB. Both capped.

src/plugins/score-plugin-js/JS/Qml/QmlObjects.hpp
jcelerier: Done.

PR #2252, opened 2026-09-01
avnd: consume impulses on the GFX node path like the CPU path does

jcelerier:
## Problem

An Avendish texture/GPU node with a `halp::val_port<"Bang", std::optional<halp::impulse>>` input sees the impulse **on every tick** after the first one: the GFX path (`CpuFilterNode` / `GpuNode` / `GpuComputeNode`) applies controls from the execution engine's `score::gfx::Message` through `GpuProcessIns` and never resets them, whereas the CPU path clears one-shot values in `oscr::process_after_run` after the processor ran. `GpuProcessIns` also deduplicates messages by value, which is right for stateful controls but would drop a second impulse equal to the previous one.

Found while wiring the `Trigger` / `Build` impulses of the StreamDiffusion node (score-addon-librediffusion): one press of `Build` restarted the engine build every tick, forever.

## Fix (mirrors the CPU path)

- `GpuProcessIns`: optional-valued ports are events — applied once per execution message (the same message is handed to the renderer several times between ticks: `init`, `update`, every frame), never deduplicated by value.
- `GpuControlIns::clearControlIn` resets those ports after the processor ran; called from the three renderers after `operator()` / `dispatch()` / the render pass, next to `processControlOut`.

Stateful controls are untouched (same dedup as before).

## Validation

- Windows, ossia-sdk clang 22, CI flags (`-DOSSIA_SDK -DCMAKE_UNITY_BUILD=1 -DOSSIA_STATIC_EXPORT=1 -DSCORE_DEPLOYMENT_BUILD=1`): `score_plugin_avnd` and `score_addon_librediffusion` build clean.
- Behaviour traced against `gfx_exec_node::run` (an impulse inlet's slot is filled only on the tick it arrives) and `oscr::process_after_run`.

https://claude.ai/code/session_01SV6BNsGfa18QxBcaaWoe7m

### DIFF
--- a/src/plugins/score-plugin-avnd/Crousti/CpuAnalysisNode.hpp
+++ b/src/plugins/score-plugin-avnd/Crousti/CpuAnalysisNode.hpp
@@ -153,6 +153,7 @@ struct GfxRenderer<Node_T> final
     // Run the processor
     if_possible(state->runInitialPasses(renderer, commands, res, edge));
     if_possible((*state)());
+    parent.clearControlIn(*state);
 
     // Copy the data to the model node
     parent.processControlOut(*this->state);
--- a/src/plugins/score-plugin-avnd/Crousti/CpuFilterNode.hpp
+++ b/src/plugins/score-plugin-avnd/Crousti/CpuFilterNode.hpp
@@ -226,6 +226,7 @@ struct GfxRenderer<Node_T> final
     // Run the processor
     if_possible(state->runInitialPasses(renderer, commands, res, edge));
     if_possible((*state)());
+    parent.clearControlIn(*state);
 
     // Upload output buffers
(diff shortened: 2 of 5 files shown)


jcelerier:
Follow-up commit 0c3d845f30 adds the same `clearControlIn` to `CpuAnalysisNode.hpp` (the texture-in / value-out renderer, e.g. FastVLM), which the first commit missed.

One known asymmetry left as-is: on `CustomGpuNode` / `GpuComputeNode` the controls are applied in `runInitialPasses` / `runCompute`, i.e. after the node's `update()` ran, and are cleared at the end of the same frame — so an optional-valued input is observable by `dispatch()` but never by `update()`. No node in score or the avendish examples uses an optional-valued input on those two paths today; if one appears, applying the controls in `update()` (as `CpuFilterNode` does) is the consistent fix.

PR #2275, opened 2026-09-13
cmake: make the deployment feature check able to fail

jcelerier:
`cmake/ScoreFeatureCheck.cmake` is meant to abort a `SCORE_DEPLOYMENT_BUILD` that is missing any of ~35 required plug-ins and features. **It never has**, for four independent reasons — any one of which alone would have been enough.

### 1. `SCORE_FEATURES_LIST` was never populated

All 41 producers do a plain `list(APPEND SCORE_FEATURES_LIST x)` inside `src/plugins/*/CMakeLists.txt`. That creates a *directory-scoped* variable shadowing the `CACHE INTERNAL` one initialised at `CMakeLists.txt:295`, so the top-level check saw an empty list. `SCORE_PLUGINS_LIST` was fine because `setup_score_plugin` correctly re-`set(... CACHE INTERNAL)`s it; `score_declare_feature()` now does the same.

### 2. `IN_LIST` was given the list's contents, not its name

```cmake
if(NOT "${name}" IN_LIST "${SCORE_PLUGINS_LIST}")
```

`IN_LIST` takes the **name** of a list variable. CMake looked up a variable named after the whole expanded list, found nothing, and considered every entry missing.

### 3. No `PARENT_SCOPE`

`list(APPEND SCORE_MISSING_PLUGINS ...)` inside `function()` died with the function frame — which is what swallowed (2)'s false positives and left the check silently passing.

### 4. Name-space mismatch

The asserts spell plug-ins `score-plugin-gfx` while `SCORE_PLUGINS_LIST` holds target names `score_plugin_gfx`.

## Measured before proposing it

Re-ran the configure of the *actual* Linux AppImage release build with the fix: it **passes**, and the lists are genuinely populated — 35 features and 65 plug-ins. So the release has everything it asserts; the guard simply starts doing its job.

**It immediately caught a real regression.** The flatpak build now fails with:

```
Deployment build is missing the following features: analysis_kfr
```

`-DSCORE_USE_SYSTEM_LIBRARIES=1` makes `3rdparty/libossia.cmake:29-40` pick FFTW and force `OSSIA_ENABLE_KFR OFF`, so `score-plugin-analysis` never declares `analysis_kfr`. That assert is verbatim in v3.7.1 — **the flatpak has been shipping without KFR since at least then**, unnoticed. The same trap catches the GCC rows of `ubuntu-builds.yaml` and both nix definitions, which pass `SCORE_DEPLOYMENT_BUILD=1` together with `SCORE_USE_SYSTEM_LIBRARIES=1`.

## Expect packaging fallout — that is the point

In a clean Arch chroot the check now fires in two passes:

- `missing plug-ins: score-plugin-faust;score-plugin-ysfx;score-plugin-lv2`
- `missing features: sdl;protocol_joystick;pipewire`

All three AUR PKGBUILDs pass `-DSCORE_DEPLOYMENT_BUILD=1` while keeping those in `optdepends`, which makepkg does not install in a clean chroot — so they have been silently shipping a crippled package. Fixing that is a packaging decision, not part of this PR.

Also worth noting: `score_assert_plugin(score-plugin-jit)` is guarded by `CMAKE_SYSTEM_PROCESSOR MATCHES "AMD64"`, which never matches on Linux, so JIT is effectively unasserted there.


jcelerier:

PR #2276, opened 2026-09-13
ci: let the AppImage build run outside the CI runner's assumptions

jcelerier:
Three small things in `ci/appimage.build.sh`, all found while reproducing the official AppImage build for the 3.9 release validation on a developer machine. None changes what CI does today; they stop the script from breaking the moment it runs anywhere else.

### `BUILD_FOLDER` was hardcoded

`export BUILD_FOLDER=/tmp/build` cannot work where `/tmp` is a small tmpfs — here it is a 32 GB tmpfs with 2.8 GB free. Now `${BUILD_FOLDER:-/tmp/build}`, so a caller can relocate it.

### …and then ignored again

The `appimagetool` invocation passed the literal `/tmp/build/score.AppDir` while every other line goes through the `build` symlink. So relocating the build folder got all the way through compiling score and then died in packaging with

```
Error: no such file or directory: /tmp/build/score.AppDir
```

### The chown could not work, and hid that it hadn't

```sh
sudo chown -R "$(whoami)" "$BUILD_FOLDER" 2>/dev/null || chown -R "$(whoami)" "$BUILD_FOLDER" 2>/dev/null || true
```

Under rootless docker (the self-hosted runner) the files are already ours; on a GitHub-hosted runner passwordless `sudo` works. With **rootful docker and a password-protected sudo** — an ordinary developer box — both arms fail and `|| true` swallows it. The six following `cp`/`sed` commands into the AppDir then all fail with `Permission denied`, the script keeps going, and the first visible symptom is `appimagetool` reporting `Could not symlink .DirIcon` much later.

Doing the chown from a container needs no sudo in either mode — being in the `docker` group is already root-equivalent — and behaves identically under rootless docker. It is also no longer allowed to fail quietly.

Verified by building the 3.9 AppImage end to end with these changes: 2576 targets including all 27 in-tree add-ons, `--version` → `score 3.9.0`.

### Adjacent, not changed here

`cmake/Deployment/Linux/AppImage/Dockerfile.llvm` clones `jcelerier/cninja` unpinned at image-build time. Nothing is broken today (cninja has not moved since 2025-03-02) and the self-hosted runner's warm layer cache hides it, but a cold cache on a new runner would silently pick up whatever master is at that moment. Worth pinning a tag or sha.


jcelerier:

Issue #10, opened 2015-05-26
state machine crashes when background is selected


nvuaille:
On 'dev' branch.
- launch i-score.
- click outside the main box (e.g.  in the area at the bottom). The border should disappear.
- then try to change edition mode (e.g. from select to move)
- SEGFAULT

0   QMetaObject::activate(QObject_, int, int, void__)   /home/nicolas/Qt/5.4/gcc_64/lib/libQt5Core.so.5     0x7ffff6518d66  
1   ScenarioStateMachine::exitState moc_ScenarioStateMachine.cpp    230 0x7fffd84a0e8d  
2   ScenarioStateMachine::changeTool    ScenarioStateMachine.cpp    155 0x7fffd840b361  
3   ScenarioControl::<lambda()>::operator()(void) const ScenarioControl.cpp 324 0x7fffd8423154  
4   QtPrivate::FunctorCallQtPrivate::IndexesList<, QtPrivate::List<>, void, ScenarioControl::makeToolbars()::<lambda()>>::call(ScenarioControl::<lambda()>, void *_)  qobjectdefs_impl.h  494 0x7fffd8426d87  
5   QtPrivate::FunctorScenarioControl::makeToolbars()::<lambda(), 0>::callQtPrivate::List<, void>(ScenarioControl::<lambda()> &, void _, void *_)   qobjectdefs_impl.h  551 0x7fffd8426a1a  
6   QtPrivate::QFunctorSlotObjectScenarioControl::makeToolbars()::<lambda(), 0, QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase _, QObject *, void *_, bool _) qobject_impl.h  192 0x7fffd8426595  
7   QMetaObject::activate(QObject_, int, int, void**)   /home/nicolas/Qt/5.4/gcc_64/lib/libQt5Core.so.5     0x7ffff65191d5  
8   QAction::triggered(bool)    /home/nicolas/Qt/5.4/gcc_64/lib/libQt5Widgets.so.5      0x7ffff73826b2  
9   QAction::activate(QAction::ActionEvent) /home/nicolas/Qt/5.4/gcc_64/lib/libQt5Widgets.so.5      0x7ffff73844d7  
10  ??  /home/nicolas/Qt/5.4/gcc_64/lib/libQt5Widgets.so.5      0x7ffff748e8c5  
[...]

PR #2277, opened 2026-09-13
flatpak: reconcile the in-repo manifest with flathub

jcelerier:
The in-repo flatpak and `flathub/io.ossia.score` have drifted in **both** directions since flathub PR #10 established the design that the `modules/` yamls are shared and only the `sources:` block differs. This brings them back together, on top of a full build. `modules/` is now name-for-name identical to flathub's, and the manifest diff against the prepared flathub version is exactly the `sources:` block and nothing else.

Built and run: `ossia-score --version` → `score 3.9.0`; `--no-gui` with `QT_QPA_PLATFORM=offscreen` starts, registers QML types, loads Pd, and stays up.

## Taken from flathub (we were behind)

| | before | after |
|---|---|---|
| `org.freedesktop.LinuxAudio.Plugins` | `24.08` | `25.08` |
| `merge-dirs` | `vst;vst3;lv2` | `vst;vst3;lv2;clap` |
| plug-in paths | `VST_PATH`/`VST3_PATH` via `finish-args` | all four exported from the run script, incl. `CLAP_PATH` |
| C compiler | `CXX` only | `CC` too |

**CLAP plug-ins were not merged into the extension directory at all**, and `CLAP_PATH` was never set — so the flatpak could not see them. flathub's PR #9 ("force clang as C compiler for uniformity") was merged there and never mirrored here, so C sources were building with the runtime's default compiler.

## Modules the runtime already provides

Dropped `abseil-cpp`, `avahi`, `libsamplerate`, and an `fmtlib.yaml` **no manifest ever referenced**. The 6.10 runtime ships absl 20250512, avahi-client 0.8, samplerate 0.2.2 and fmt 11.2.0 — this is what flathub PR #20 already did.

Verified in the built sandbox rather than assumed: `ldd` over the app reports **181 dependencies, 0 "not found"**. `libsamplerate.so.0`, `libfmt.so.11` and 17 `libabsl_*` all resolve from the Platform runtime. `libre2` correctly still comes from `/app`. avahi turns out not to be linked at all — `dnssd` comes from the vendored `3rdparty/Servus` copy.

## Fixes found by building it

- **`--socket=x11` together with `--socket=wayland` is a hard flathub lint error** (`finish-args-contains-both-x11-and-wayland`). Changed to `--socket=fallback-x11`, which is the documented pairing; the linter then exits 0. ⚠️ **This is a real behaviour change** — under Wayland score no longer gets an X11 socket, so anything assuming XCB (suil plug-in UI embedding, gfx window handles) takes a different path. The offscreen test says nothing about that; **worth one manual Wayland run before tagging**, or drop `--socket=wayland` instead.
- **`-DOSSIA_ENABLE_KFR=1`**: `SCORE_USE_SYSTEM_LIBRARIES` makes `3rdparty/libossia.cmake:29-40` pick FFTW and force KFR off, and a deployment build asserts `analysis_kfr`. That assert is verbatim in v3.7.1 — **the flatpak has been shipping without KFR since at least then**; it only surfaces now because #2275 makes the check able to fail.
- `shared-modules` was pinned to a **2024-06-25** commit.
- The CI container was still `kde-6.9` while the manifest asks for the 6.10 runtime. `kde-6.10` exists and carries both `linux/amd64` and `linux/arm64`, so it covers both matrix legs.

## Deliberately left alone, but you should look

`-DCMAKE_CXX_FLAGS="-Wl,--strip-all"` is not cosmetic: `ossia-score.bin` is stripped and there is **no `*ossia*` file anywhere in `lib/debug`**, so the 55 MB Debug extension carries debug info for the bundled dependencies only. Every flatpak crash report will be bare addresses. It is not in flathub master either. I have not removed it — that is your call.

Also unproven: llvm21 is not *required*. `org.freedesktop.Sdk.Extension.llvm20/x86_64/25.08` exists alongside llvm21, so runtime 6.10 does not force the move, and no version gate in score asks for 21 (`CMakeLists.txt:181` is dead — `:188` sets `CMAKE_CXX_SCAN_FOR_MODULES 0` unconditionally seven lines later). Keeping 21 for toolchain alignment is fine; I did not build with 20, so "clang 20 still works" is untested.


jcelerier:

PR #2284, opened 2026-09-16
gris: VBAP and MBAP spatialisation from SpatGRIS as a native process

edumeneses:
This draft brings the two amplitude-panning algorithms of SpatGRIS, the spatialisation server developed by the Groupe de recherche en immersion spatiale (GRIS, Université de Montréal), into score as a single native audio process. The goal is to spatialise sources over large loudspeaker arrays directly on the timeline (we run a 93-speaker dome and a 97-speaker cube at the Société des arts technologiques), with source positions arriving as ordinary control inlets rather than through an external OSC-driven application.

## What it adds

- `src/plugins/score-plugin-gris`, an in-tree plugin providing one process: N audio inputs, M audio outputs, both Vector Base Amplitude Panning (VBAP) and Matrix-Based Amplitude Panning (MBAP) resident, with the algorithm selectable per source at run time (the shape SpatGRIS itself uses in its hybrid mode).
- Per-source inlets for position (`vec3f`), azimuth span, zenith span and algorithm; a global gain-interpolation control; a source count that is changed through an undoable command, so cables and automations on existing sources survive when the count grows.
- A speaker-setup port whose editor is a table (following `Gfx::Images::ImageListChooser`): it reads the three speaker-setup XML variants SpatGRIS has written over the years (legacy `SPEAKER_SETUP`, the intermediate `SpeakerSetup/Ring/Speaker` form, and the current v4 ValueTree with nested `SPEAKER_GROUP` and yaw/pitch/roll), and writes the current one back, so setups round-trip with SpatGRIS.
- Gain smoothing inside the audio loop, as upstream does it: a per-(source, speaker) ramp interleaved with the multiply, linear or first-order, so moving sources do not zipper.

## What it deliberately leaves out

Mute/solo, direct outs, master gain, per-speaker high-pass, recording and the OSC input are not ported; score's mixer and device tree already provide them. HRTF/binaural and stereo reduction are out of scope because they cannot be expressed as a gain matrix (and are being rewritten upstream). The applier is factored as a free function over a gain matrix, so a gain vector computed elsewhere (DBAP, GBAP, an external inlet) could be fed through the same interpolation later without a refactor.

## Provenance and licence

The gain computers in `Gris/Algo/` are ports of `sg_vbap.cpp` and `sg_mbap.cpp` from AlgoGRIS (Gaël Lane Lépine, Samuel Béland, Olivier Bélanger and Nicolas Masson; GPLv3, as score is), rewritten without JUCE: only `SpeakerSetupIO` touches Qt, through `QXmlStreamReader`. The algorithms were verified against a standalone harness: the Satosphère dome, the SAT studio cube, the binaural default and a 16-speaker ring all parse, spatialise, and round-trip through the writer, and a source placed on a speaker comes out loudest on that speaker (unity gain for VBAP).

## Status and open questions

The `Gris/Algo` layer is tested; the score-facing layer (model, executor, inlet, inspector) is compiling for the first time as this draft is opened, so expect fix-ups. Two things I would like your view on, @jcelerier:

1. The documentation describes a `matrix` process and a GBAP process, with working example scores, but I could not find either in master or in avendish. If they exist somewhere, the applier here should probably become that node instead.
2. Whether this belongs in-tree or as an addon, and whether a thin Avendish object over the same `Gris/Algo` engine (for a standalone or plug-in build) would be welcome alongside it.

https://claude.ai/code/session_01341ZRRMM5j8eLPieQ53Np8


jcelerier:
Is it possible to put this in this repo and continue this PR  instead - the repo is called "GBAP" but is more realistically "spatialization tools" :  https://github.com/ossia/GBAP/pull/33

jcelerier:
Also for vbap did you check against the Faust VBAP that score already supports ? (need to install abclib from the package manager)

edumeneses:
The plan was to use SpatGRIS' implementation to switch on the fly between VBAP/MBAP. I would like to keep our current workflow and also take advantage of the speaker setup that needs to be implemented for MBAP.

On https://github.com/ossia/GBAP, we can move there, but if we're doing all sound spatialization there, can we rename it?

jcelerier:
there's also http://github.com/ossia/score-addon-spat which uses https://github.com/leomccormack/Spatial_Audio_Framework ; maybe all of this can be merged and finalized :D 

> take advantage of the speaker setup that needs to be implemented for MBAP.

that's what GBAP#33 adds :D

edumeneses:
Ok, but I am proposing an audio process. The workflow I imagine is to have the "audio_spatializator" process at the end of the audio chain right before sending the audio objects to the score output channels.

The GBAP process doesn't currently work with audio. I would be more inclined to merge what I did here with [score-addon-spat](https://github.com/ossia/score-addon-spat) and keep old GBAP as a legacy process. We can then maintain a single spat addon that benefit from AlgoGRIS future developments. 

Only caveat is that for now there are diffs on AlgoGRIS to make it JUCE-free for my process. Ideally, AlgoGRIS should be JUCE-free by default and we could consume it directly. The cleaning process, if done on our end, will require extra work when AlgoGRIS gets an update we want to downstream.

BTW, the current version of GBAP outputs only zeros on the array (tested on my Ubuntu 26.04 using latest score build at the time of this message). I would like to confirm with another user testing it before open an issue.

jcelerier:
@bltzr can you check for GBAP?

PR #2291, opened 2026-09-18
gfx: pin the default surface format so Qt Quick's RHI is desktop GL, not GLES

ogauthiersat:
Follow-up to 7e80cdb, which fixed this where score owns the QRhi. It does not
reach the case where Qt Quick owns it.

Rebased onto dc6d727.

## Symptom

A custom app (`--ui`) whose QML renders a score node through `UI.TextureSource`
draws nothing from score, forever, while the rest of the QML scene is fine:

```
score.gfx: RHI device: backend=OpenGL device="NVIDIA ... RTX 3070 ... OpenGL ES 3.2 NVIDIA 580.126.09"
No GLSL shader code found (versions tried: QList(Version(320 GlslEs), Version(310 GlslEs),
  Version(300 GlslEs), Version(100 GlslEs))) in baked shader QShader(... ShaderKey(1 Version(460)))
Warning! Pipeline not created
Could not find matching pipeline for draw
```

The window is alive and the app looks healthy, so this reads as "score renders
nothing in QML" rather than as a graphics-init failure.

## Cause

`score::gfx` bakes desktop GLSL only — `importRenderState()` takes the shader
version from `score::GLCapabilities`, which probes with `QSurfaceFormat::OpenGL`
+ `CoreProfile` explicitly requested and reports 460. Nothing forced the contexts
to match:

1. `QSurfaceFormat::defaultFormat()` is never pinned on desktop Linux.
   `setupDefaultOpenGLFormat()` is a no-op there, and the probe in
   `setup_opengl()` that would call `setDefaultFormat()` is gated on
   `appSettings.opengl` — which `ApplicationSettings::parse()` forces to `false`
   unless `--opengl` is passed. So `renderableType` reaches the driver as
   `DefaultRenderableType`.
2. `setup_x11()` puts every X11 launch on `QT_XCB_GL_INTEGRATION=xcb_egl`, GLX
   being unable to import dma-bufs. An EGL that serves both APIs — NVIDIA
   advertises `EGL client APIs: OpenGL_ES OpenGL` — then answers with **GLES**.
3. `QSGRhiSupport::createRhi()` builds both `QRhiGles2InitParams::format` and
   `newFallbackSurface()` from `QWindow::requestedFormat()`, so Qt Quick's QRhi
   inherits that GLES context. Qt Quick does not care, its own shaders are baked
   for both.
4. `JS::TextureSource` hands that QRhi to a `PreviewNode`. QRhiGles2 looks for
   GLSL ES variants score never baked → no pipeline → every draw skipped.

A per-window fix does not cover this. A custom app whose QML root is an
`ApplicationWindow` creates and shows its own window;
`JS::ApplicationPlugin::afterStartup()` never touches it — its
`qobject_cast<QQuickItem*>` fails and it falls through. The process-wide default
format is the only input left.

## Commits

**`gfx: pin the default surface format when Qt Quick runs on OpenGL`**

`score::pinDefaultOpenGLFormat()` applies `GLCapabilities::setupFormat()` to
`QSurfaceFormat::defaultFormat()`, called from `update_qtquick_graphics_api()`
in the `OpenGL` case — where score commits Qt Quick to a backend, and which
`::bind()` invokes once at `Gfx::Settings::Model` construction, before any window
exists. Using the same `GLCapabilities` for the format and for the baked shader
version is what keeps them from diverging.

It declines twice:

- on `minimal` / `offscreen` / `vnc` / `wasm`, the list `setup_opengl()` already
  skips, since the probe creates a context;
- when the probe found no usable context — the path 4cb7812356 added — because it
  then reports the format it *asked for*, so pinning would put that fallback, a
  `CoreProfile` 2.0, on every window in the process.

The probe records that distinction in `usableContext`; it stays private behind a
`glCapabilities()` accessor rather than widening `GLCapabilities`, so the public
header gains only the function declaration.

**`gfx: set the GL renderable type before creating the KMS fallback surface`**

`KmsOutputNode::createOutput()` still had the pre-7e80cdb shape:
`newFallbackSurface()` with no argument, then `caps.setupFormat()` afterwards.
Latent rather than observed — it needs a DRM scanout output, which this machine
does not have.

**`gfx: name the renderer the GL capability probe ran on`**

The probe line reported what it got but not who answered, and the shader version
every `score::gfx` pipeline is baked against comes from that context. `RenderList`
prints the device for the RHI it ends up with, not for the probe that decided the
bake. Appends `GL_RENDERER`, read while the context is current. Diagnostic only —
nothing branches on it.

```
 src/lib/score/gfx/OpenGL.cpp                             | 43 ++++++++++++++++--
 src/lib/score/gfx/OpenGL.hpp                             |  3 ++
 src/plugins/score-plugin-gfx/Gfx/Graph/KmsOutputNode.cpp |  3 +-
 src/plugins/score-plugin-gfx/Gfx/Settings/Model.cpp      |  2 +
 4 files changed, 51 insertions(+), 4 deletions(-)
```

## Verification

NVIDIA 580.126.09 / RTX 3070, X11 under `xcb_egl`, Qt 6.8.2. Custom app launched
with `--autoplay --no-restore --ui Main.qml app.score`, whose QML feeds
`UI.TextureSource` into a Qt Quick 3D `CustomMaterial`:

| | before | after |
|---|---|---|
| RHI device | `OpenGL ES 3.2 NVIDIA 580.126.09` | `4.6.0 NVIDIA 580.126.09` |
| `No GLSL shader code found` | repeating | 0 |
| `Pipeline not created` | repeating | 0 |
| `Could not find matching pipeline` | repeating | 0 |
| textured geometry | absent | renders |

The mechanism was also isolated ahead of the fix: a standalone QQuickWindow under
`QT_XCB_GL_INTEGRATION=xcb_egl` comes up `OpenGLES` / `OpenGL ES 3.2` unpinned and
`OpenGL` / `4.6.0` pinned, on the same machine.

Other configurations:

- normal GUI run (no flags) — starts clean,
  `Available GL context: 4 6 460 QSurfaceFormat::OpenGL "NVIDIA GeForce RTX 3070/PCIe/SSE2"`,
  no GL errors;
- Mesa llvmpipe forced through EGL
  (`__EGL_VENDOR_LIBRARY_FILENAMES=.../50_mesa.json LIBGL_ALWAYS_SOFTWARE=1`) —
  pins `4 5 450`, RHI comes up `Mesa llvmpipe ... 4.5 (Core Profile)`, zero shader
  or pipeline warnings, no crash. llvmpipe advertises 4.5 core, so the pin asks
  for something it supports; `main.cpp`'s llvmpipe bail also disables the
  `QOpenGLWidget` viewport, which is a separate concern from the surface format;
- `QT_QPA_PLATFORM=offscreen` — probe never runs, no crash;
- `ReadbackTester` aborts identically with and without the patch (pre-existing,
  confirmed by stashing).

## Disclosure


src/lib/score/gfx/OpenGL.cpp
```diff
@@ -45,15 +48,27 @@ struct GLCapabilitiesResult
...
+      {
+        const auto renderer = QString::fromUtf8(r);
+        softwareRasterizer = renderer.contains("llvmpipe", Qt::CaseInsensitive)
+                             || renderer.contains("softpipe", Qt::CaseInsensitive)
+                             || renderer.contains("swrast", Qt::CaseInsensitive);
+      }
+    }
 #endif
-    qDebug() << "Available GL context: " << major << minor << shaderVersion << type;
+    qDebug() << "Available GL context: " << major << minor << shaderVersion << type
```
jcelerier, line 70: I think maybe directly printing the renderer string would be good ?

src/lib/score/gfx/OpenGL.cpp
```diff
@@ -177,4 +193,22 @@ void GLCapabilities::setupFormat(QSurfaceFormat& fmt)
...
+void pinDefaultOpenGLFormat() noexcept
+{
+#ifndef QT_NO_OPENGL
+  // Same list setup_opengl() skips: no GL, or a crash in QOffscreenSurface::create.
+  const auto plat = QGuiApplication::platformName();
+  if(plat == "minimal" || plat == "offscreen" || plat == "vnc" || plat == "wasm")
+    return;
+
+  GLCapabilities caps;
+  if(caps.softwareRasterizer)
```
jcelerier, line 206: I wouldn't skip that part with setDefaultFormat. The actual problem the software rasterization avoidance was was a crash of llvmpipe when trying to open a context before we had a window (but, conversely, with Qt Quick or the main graphics view's OpenGL mode, we need to have an operational GL context setup by the moment we show it). But once we have a window (which is usually the case in OpenGL.cpp) then I don't remember any crash occuring.

src/lib/score/gfx/OpenGL.cpp
jcelerier: e.g. more than a few times I lost quite some time chasing a bug because the "GPU" that was picked was a random HDMI dongle or the intel gpu instead of nvidia

src/plugins/score-plugin-gfx/Gfx/Graph/KmsOutputNode.cpp
```diff
@@ -323,11 +323,12 @@ void KmsOutputNode::createOutput(OutputConfiguration conf)
   d->state->samples = 1;
 
   QRhiGles2InitParams params;
-  d->state->surface = QRhiGles2InitParams::newFallbackSurface();
```
jcelerier, line 326: Could you try this one ? it's to be used when you have no X server / wayland at all

jcelerier:
also we'd need to run a test on a rpi: there, the default Qt build that comes with raspbian is built with GLES enabled and desktop GL disabled ; likewise for the WASM builds where it's only GLES

src/lib/score/gfx/OpenGL.cpp
ogauthiersat: I added logging of the renderer. I tested it by forcing software renderer :

```
__EGL_VENDOR_LIBRARY_FILENAMES=/usr/share/glvnd/egl_vendor.d/50_mesa.json LIBGL_ALWAYS_SOFTWARE=1 build/ossia-score
```

```
Debug: Available GL context:  4 5 450 QSurfaceFormat::OpenGL "llvmpipe (LLVM 19.1.7, 256 bits)" (OpenGL:88)
```

While hardware renderer logs :

```
Debug: Available GL context:  4 6 460 QSurfaceFormat::OpenGL "NVIDIA GeForce RTX 3070/PCIe/SSE2" (OpenGL:88)
```

src/lib/score/gfx/OpenGL.cpp
ogauthiersat: I removed the software rasterizer guard and kept the default format pin if probe returns a usable context.

ogauthiersat:
> ` [3588736](/ossia/score/pull/2291/commits/358873685c7c422eaef18e4604499c3ad4f98ec8)`

I tested the wasm CI build (6b481e1 from 2026-09-18). It appears to work, here's the logging I got with an ISF Shader process.

```
10:41:41.470 Debug: Available GL context:  3 0 300 QSurfaceFormat::OpenGLES  (:0) ossia-score.js:1:1136049
10:41:41.665 Debug: score.gfx: RHI device: backend=OpenGL device="Mozilla NVIDIA GeForce GTX 980, or similar OpenGL ES 3.0 (WebGL 2.0)" vendorId=0x0 deviceId=0x0 deviceType=unknown (:0) ossia-score.js:1:1136049
10:41:41.666 Debug: score.gfx: RHI indirect caps: drawIndirect=0 multi=0 count=0 dispatchIndirect=0 baseInstance=0 instIdxInclBase=0 (:0)
```

I'll give the arm64 appimage a test on Raspberry and see if I can get wayland and eglfs/kms working.

ogauthiersat:
I tested aarch64 appimage build on raspios trixie 2026-09-15 on a Raspberry Pi 4.

I tried with the default Wayland setup, with X11 (QT_XCB_GL_INTEGRATION=xcb_glx and xcb_egl) and with eglfs/KMS.

UI renders correctly in all cases but I could not get OpenGLES to load a context, ISF Shader process preview always renders yellow.

I tested this branch build 3588736 and master build c92fc08.

glxinfo :

```
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Broadcom (0x14e4)
    Device: V3D 4.2.14.0 (0xffffffff)
    Version: 26.2.2
    Accelerated: yes
    Video memory: 3795MB
    Unified memory: yes
    Preferred profile: core (0x1)
    Max core profile version: 3.1
    Max compat profile version: 3.1
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.1
OpenGL vendor string: Broadcom
OpenGL renderer string: V3D 4.2.14.0
```
KMS/DRM rendering without Wayland/X11 works, I simply stopped lightdm (sudo systemctl stop lightdm) then launched the appimage via SSH. The mouse and keyboard worked out of the box.

Here's the debug log : 

```
QEGLPlatformContext: Failed to create context: 3009
score: no usable OpenGL context for capability probing (surface true ); keeping the requested format 3 2 320
Available GL context:  3 2 320 QSurfaceFormat::OpenGLES ""

Warning: QEGLPlatformContext: Failed to create context: 3009 (:0)
Warning: QRhiGles2: Failed to create temporary context (:0)
Warning: QEGLPlatformContext: Failed to create context: 3009 (:0)
Warning: QRhiGles2: Failed to create context (:0)
Critical: createRenderState: QRhi::create(OpenGLES2) FAILED. This output will never render. (:0)
Warning: score::gfx: NULL RHI BACKEND — no GPU backend could be created, so nothing will actually be rendered. Under QT_QPA_PLATFORM=offscreen this is expected: Qt's offscreen integration provides OpenGL only through GLX, so with no X server there is no GL at all. Use a real X server (Xvfb) with QT_QPA_PLATFORM=xcb for a run that renders. (:0)
Debug: score.gfx: RHI device: backend=Null device="Null" vendorId=0x0 deviceId=0x0 deviceType=unknown (:0)
Debug: score.gfx: RHI indirect caps: drawIndirect=1 multi=1 count=1 dispatchIndirect=1 baseInstance=1 instIdxInclBase=1 (:0)
```

jcelerier:
thanks for all the testing!

src/lib/score/gfx/OpenGL.cpp
jcelerier: super thanks!!

PR #2294, opened 2026-09-20
gfx: cross-backend CSF feedback, ISF culling, depth lifetime and leak fixes

jcelerier:
Graphics-engine work across the CSF / ISF / raw-raster paths, validated on
OpenGL, Vulkan and Metal on real displays. 52 commits, 75 files.

## What this fixes

**CSF geometry feedback loops.** A compute pass that feeds its own geometry
back did not advance on OpenGL when the loop had an even number of nodes: QRhi
inserts no compute-to-compute barrier across pass boundaries there, so a CSF
pass now emits its own. All three backends agree afterwards. A `read_write`
snapshot was also being allocated one element wide and never resized, and a
feedback receiver could adopt the shared upstream handle and undo its own swap.

**Cross-backend `CULL_MODE` for ISF.** The ISF vertex epilogue mirrors Y under
SPIRV/HLSL/MSL while `isf_vertShaderInit` already applies
`clipSpaceCorrMatrix`. The two cancel for position, so image orientation was
always correct, but the mirror reverses window-space winding — so the same
geometry rasterized CCW on OpenGL and CW on Vulkan and Metal, and `CULL_MODE`
meant the opposite thing depending on backend. Measured over four shader shapes
crossed with three cull states on three backends: 8 of 12 rows disagreed
before, 0 after, with every uncalled frame RMSE 0 against its pre-fix frame.

**Depth.** Sink depth attachments were never released on the falling edge, so a
sink held its attachments for the whole session; depth invalidation is now
symmetric. Reverse-Z defaults are applied consistently, including a splat path
that cleared to 0.0 but still compared `Less`.

**Leaks and lifetimes.** The per-frame `QRhiResourceUpdateBatch` leaked on an
exception path; pipelines leaked their bindings on six failure paths; a
swap-chain guard could fire without a frame in flight.

**ISF parsing.** `main()` renaming matched inside comments and `#if 0` blocks;
it now detects the entry point on a comment-stripped copy and lets the
preprocessor do the rename. A mandatory-`ACCESS` throw fired before the
forward-declaration case could clear it.

## Tests and harness

New coverage for the feedback loop (including the even-length case that started
this), for `read_write` state, and for gather in/out sizing. The integration
harness no longer depends on one machine's folder layout, device indices or a
guessed Vulkan physical-device index, and skips cleanly (exit 77) where a
device family is absent — it had been reporting stale PNGs from previous runs
as passes.

## Not included

The corpus `gather` declarations live in `score-corpus` and are not part of
this repo. Six pre-existing test failures under ctest are an asynchronous
library scan racing Catch2 teardown (`RecursiveWatch::scanAsync` ->
`Faust::LibraryHandler::scanPath`), proven from a symbolised core dump; nothing
here touches that code.

https://claude.ai/code/session_01JeL19ofVgPe9u9q7J1V9ex


src/plugins/score-plugin-gfx/Gfx/Graph/RenderedCSFNode.cpp
```diff
@@ -3734,6 +3749,68 @@ void RenderedCSFNode::buildComputeSrbBindings(
...
+                ssbo.read_buffer_is_snapshot = false;
+              }
+              auto* snap = rhi.newBuffer(
+                  QRhiBuffer::Static,
+                  QRhiBuffer::StorageBuffer | QRhiBuffer::VertexBuffer,
+                  ssbo.buffer->size());
+              snap->setName(QByteArray("CSF_GeomSnap_") + req.name.c_str());
+              if(snap->create())
+              {
+                QByteArray zero((int)ssbo.buffer->size(), 0);
```
jcelerier: !

src/plugins/score-plugin-gfx/3rdparty/libisf/src/isf.cpp
```diff
@@ -3992,7 +4061,22 @@ void parser::parse_isf()
 
   // Add the actual vert / frag code
   if(!simpleVS)
-    m_vertex += m_sourceVertex;
+  {
+    // Detect the entry point on a comment-stripped copy, but let the
+    // preprocessor do the rename: a textual replace also hits main() in a
+    // comment, a prototype or an #if 0 block, and renames the wrong one.
+    if(hasEntryPoint(m_sourceVertex))
```
jcelerier: either a vertex shader has an entry point or it's not provided by the user. no point in handling other cases

src/lib/score/gfx/OpenGL.cpp
```diff
@@ -45,13 +45,32 @@ struct GLCapabilitiesResult
...
+    // means querying an unusable context. Keep the conservative default
+    // instead: a wrong-but-safe shader version degrades one feature, a
+    // context that was never created takes the process down.
+    const bool surfaceOk = surf.isValid();
+    const bool contextOk = surfaceOk && ctx.create() && ctx.makeCurrent(&surf);
+    if(!contextOk)
+    {
+      qWarning() << "score: no usable OpenGL context for capability probing"
+                 << "(surface" << surfaceOk << "); assuming GLSL 1.20";
+      major = 2;
```
jcelerier: Don't force 2.0, instead just let the default Qt surface format unless there is a good reason not to

src/plugins/score-plugin-gfx/3rdparty/libisf/src/isf.cpp
```diff
@@ -27,6 +27,39 @@ namespace isf
 {
 namespace
 {
+//! True when the source defines main() outside comments and disabled blocks.
+bool hasEntryPoint(std::string_view src) noexcept
```
jcelerier: I thought we had said we wouldn't be replacing the main and instead rely on shaders adding proper calls to
isf_vertShaderInit();
isf_vertShaderFinish();
?

src/plugins/score-plugin-gfx/Gfx/Graph/RenderedCSFNode.cpp
jcelerier: Didn't we have a static "zero" buffer that we were already using for this ?

src/plugins/score-plugin-gfx/Gfx/Graph/Utils.cpp
```diff
@@ -1179,6 +1220,27 @@ Pipeline buildPipelineWithState(
   applyPipelineState(
       *ps, state, rt.colorAttachmentCount(), depthAvailable, wantsDepthByDefault);
 
+  // The ISF vertex epilogue mirrors Y wherever the shader is not GLSL, which
```
jcelerier: what are situations where the shader is not GLSL ?

src/plugins/score-plugin-threedim/Threedim/ModelDisplay/ModelDisplayNode.cpp
```diff
@@ -117,8 +117,6 @@ vec4 v_projected = vec4(0.0, 0.0, 0.0, 1.0);
   vec3 d = viewspace.xyz;
   float r = length(d);
 
-  const float PI = 3.14159265358979323846264338327;
```
jcelerier: did you validate that this works on every shader dialect we compile to, down to GLES2, HLSL for D3D11/12, MSL, etc. ?

src/plugins/score-plugin-scenario/Dataflow/Commands/CableHelpers.cpp
```diff
@@ -386,29 +386,27 @@ void reloadPortsInNewProcess(
   // values are the point and an old one restored over them would be wrong.
   constexpr auto flags = Process::PortLoadDataFlags::ReloadValue;
 
-  // Try an optimistic matching. Type and name must match.
```
jcelerier: Explain the change in this file

PR #2296, opened 2026-09-21
gfx: fix the raw-raster MULTIVIEW binding shift, and give raw raster a camera

jcelerier:
Two faults found while diagnosing "MODEL_MATRIX in the raw raster pipeline is broken". Each is pinned by a test that was red before its fix and green after.

## 1. MULTIVIEW shifted every binding past the model UBO

`parse_raw_raster_pipeline()` emits `multiview_t` at `sampler_binding` when `MULTIVIEW >= 2` and then takes `model_ubo_binding` from the incremented counter, so the generated shader declares `model_material_t` one slot higher. `RenderedRawRasterPipelineNode` counted its own bindings from `int max_binding = 3` plus samplers plus auxiliary textures and reserved nothing for multiview — it had no multiview UBO at all, where `RenderedISFNode` and `SimpleRenderedISFNode` both allocate one. Every binding from the model UBO on was one lower than the shader asked for, and `MODEL_MATRIX` read a slot nothing was bound to.

Nothing caught it because no multiview raw-raster shader read `MODEL_MATRIX`: `syn-camera-array-faces` reads the `camera` auxiliary and synthesises its triangle from `gl_VertexIndex`.

Fixed by allocating and binding the buffer the shader declares, identity-seeded for the same reason `RenderedISFNode` seeds its own — no producer fills the per-view matrices yet, and zeros collapse every vertex to the origin.

## 2. Raw raster had no camera a shader could ask for

The prelude supplies `renderer_t`, `process_t` and `model_material_t`; the only `viewProjection` in `isf.cpp` is the MULTIVIEW block. View and projection reached a shader solely through the ScenePreprocessor's `camera` auxiliary, which the shader had to declare by name and index as a flat `vec4` array against the 240-byte std140 `CameraUBOData`. So `clipSpaceCorrMatrix * MODEL_MATRIX * position` — the idiom 33 corpus scores write — had no camera term: moving the eye from z=3 to z=12 changed **exactly 0 pixels**.

`parse_raw_raster_pipeline` now synthesises that auxiliary when the shader does not declare its own, with a typed `isf_camera_t` mirroring `CameraUBOData`, and exposes `VIEW_MATRIX` / `PROJECTION_MATRIX` / `VIEWPROJECTION_MATRIX` / `CAMERA_POSITION`.

Three things it deliberately does not do:

- **It does not touch `MODEL_MATRIX`.** A bare `MODEL_MATRIX` shader still sees no camera, because that is what `MODEL_MATRIX` means. Whether a scene `Transform3D` should land there instead of only in `per_draws` is a separate design question, left open.
- **It does not add an input.** A `uniform_input` builds a `TextureInlet` (`ISFProcess.hpp`), so declaring one would add a port to every raw-raster process and shift the inlet indices every saved document resolves its cables against. An auxiliary creates no port.
- **The accessors index through `VIEW_INDEX`**, not a scalar — face *i* must read camera *i*. A scalar resolving to camera 0 would paint all six cubemap faces identically. `VIEW_INDEX` is now defined as 0 outside multiview so one macro is correct in both.

The camera placeholder is the one auxiliary seeded with identities instead of zeros: without a ScenePreprocessor upstream there is nothing to resolve against, and an all-zero viewProjection would be strictly worse than the behaviour it replaces.

## Verification

Unit test on all three backends: **Vulkan 46/46, Metal 46/46, OpenGL 15/15**. OpenGL's lower count is the documented gate `GfxCameraArrayFaces` uses — procedural MULTIVIEW layered raster is not renderable headless there — so its six-face oracle is skipped while its camera half runs fully (294 pixels differing).

126-score corpus sweep on all three, against the post-G1 baselines:

| backend | before | after | change |
|---|---|---|---|
| Vulkan | 84 ok / 36 blank / 6 noframe | 83 / 37 / 6 | `particles-3-new` ok→blank |
| OpenGL | 59 / 60 / 7 | 60 / 60 / 6 | `lines-full-hdr` noframe→ok |
| Metal | 61 / 58 / 7 | 61 / 58 / 7 | none |

**No new `noframe` on any backend** — the failure mode a descriptor mismatch produces, and the main risk given every raw-raster shader gains a binding.

The one Vulkan change is a flake, measured rather than assumed: `particles-3-new` re-ran 6x on this binary at sd 0.0075–0.0264, all above the 0.005 blank threshold, against a sweep outlier of 0.00076. Its four prior sweeps read 0.0097–0.0337 — every value below the 0.0598 Vulkan noise floor, so it is a near-blank score that flips on nothing.

Of the 30 `MODEL_MATRIX` corpus scores in the sweep list, materially-moved sd counts were 0 on OpenGL, 0 on Metal and 1 on Vulkan — another particle score at the noise floor. The unbound camera is a genuine no-op for shaders that do not ask for it.

## Not done

The descriptor-limit question: every raw-raster shader now carries one extra binding, and `classic_pbr_openpbr` already sits against real ceilings (90 samplers vs OpenGL's 32, per the material-binding plan). I have not verified that the extra UBO does not push a heavy shader over a descriptor-set limit on the tightest backend; the corpus cannot answer it, since those scores are already black on GL/Metal for the sampler-budget reason.

https://claude.ai/code/session_01JeL19ofVgPe9u9q7J1V9ex


jcelerier:
superseded by #2303

